PR #21, opened 2021-04-06
Fix duplicated exported id

### LINKED ISSUE #10: [Bug] Story name can conflict with imports
In the loader at:

https://github.com/storybookjs/addon-svelte-csf/blob/b0010e5d1fb3f0556c04119888d1bf49269a543e/src/parser/svelte-stories-loader.ts#L51

There's some generated code that exports each story by their name. This conflicts with imports so for example you can't have something like this (modified from the example code):

```javascript
<script>
  import { Meta, Story, Template  } from '@storybook/addon-svelte-csf';
  import Button from './Button.svelte';
</script>

<Meta title="Button" component={Button} />

<Template let:args>
  <Button {...args}/>
</Template>

<Story name="Button" args={{ rounded: true}} />
```

Since it will try to declare Button again.

I believe this could be as simple as adding a prefix to the generated export name, such as "Story" and that seems to work locally but I am not familiar enough with the code and whether this would break something else to send a pull request for it.

j3rem1e:
Closes #10 

1. A list of allocated identifiers is build from the component
2. If a generated id is in this list, then a warning is emitted, and an hash is added to the id. The id stays stable between build, but the user is encouraged to configure his own id

### DIFF
--- a/src/parser/collect-stories.js
+++ b/src/parser/collect-stories.js
@@ -21,7 +21,7 @@ const createFragment = document.createDocumentFragment
   ? () => document.createDocumentFragment()
   : () => document.createElement('div');
 
-export default (StoriesComponent, stories) => {
+export default (StoriesComponent, { stories = {}, allocatedIds }) => {
   const repositories = {
     meta: null,
     stories: [],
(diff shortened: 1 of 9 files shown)


shilman (approved):
LGTM

PR #55, opened 2022-04-17
Fix names with special char at end

### LINKED ISSUE #54: [Bug] Special characters cause error in Storybook
### Describe the bug

When I add special characters to the name of a Story, for example: `<Story name="Horizontal (Default)" args={{}} />`, the `(` and `)` symbols appear to break stories, and they can no longer be found.

### Expected behavior

I would expect to be able to include any special characters in Story names and have them rendered appropriately.

j3rem1e:
Fix #54

Id for "Name with (Parenthesis)" was "NameWithParenthesis)" which is not a valid id.

### DIFF
--- a/src/parser/extract-id.test.js
+++ b/src/parser/extract-id.test.js
@@ -4,6 +4,9 @@ describe('extract-id', () => {
   test('name with spaces', () => {
     expect(extractId({ name: 'Name with spaces' })).toBe('NameWithSpaces');
   });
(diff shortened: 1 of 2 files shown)


shilman:
:rocket: PR was released in [`v2.0.2`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v2.0.2) :rocket:

Issue #60, opened 2022-07-06
[Bug] replace `svelte-loader`


benmccann:
### Describe the bug

The Svelte team only very begrunglingly maintains `svelte-loader` and its borderline deprecated. SvelteKit is the official way to build Svelte projects and uses Vite and `vite-plugin-svelte` by default. Similarly, we'd prefer people use Vitest to Jest and `svelte-jester` to better leverage the Vite ecosystem.

### Steps to reproduce the behavior

https://github.com/storybookjs/addon-svelte-csf/blob/675be7f108f3522fcf701ad2397339cd68c67f10/package.json#L67

It's also an issue here: https://github.com/storybookjs/storybook/blob/main/app/svelte/package.json

### Expected behavior

Should not require the use of Webpack with Svelte, but make it optional and use Vite by default. An example of Svelte, Storybook, and Vite can be found here: https://github.com/michaelwooley/storybook-experimental-vite

Maybe Storybook 7 will fix this

### Screenshots and/or logs

N/A

### Environment

Latest code

### Additional context

SvelteKit projects are now created with a `vite.config.js` by default in part to make setting up Storybook with `builder-vite` more seamless

PR #62, opened 2022-07-06
Make `svelte-loader` optional

benmccann:
Closes https://github.com/storybookjs/addon-svelte-csf/issues/60

It would be nice not to require `svelte-loader` (webpack support) for Vite projects. You need either it or `@storybook/builder-vite`, but it doesn't matter which one

I'm also removing `svelte` as optional as it seems like something you kind of need in a Svelte project though I don't really have strong feelings in whether it's listed as optional or not

Companion PR in https://github.com/storybookjs/storybook/pull/18645

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "svelte-loader": "^3.1.2"
   },
   "peerDependenciesMeta": {
-    "svelte": {
+    "svelte-loader": {
       "optional": true
     }
   },


IanVS (approved):
Makes sense to me, but it would be great to get 
@j3rem1e's opinion.

j3rem1e:
I am ok with this change.

PR #66, opened 2022-07-14
update description

benmccann:
As a newcomer to the ecosystem, I've been confused about what this plugin does for awhile. shilman clarified for me today that `addon-svelte-csf` isn't about allowing you to use csf, but allowing you to avoid using csf directly!


shilman:
:rocket: PR was released in [`v2.0.7`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v2.0.7) :rocket:

Issue #64, opened 2022-07-14
[Feature] Add Vite-Builder Support


Gaubee:
Vite-Builder is one of the office builder, pls add support.❤️

benmccann:
I think there's already support. You can see an example project here:

https://github.com/michaelwooley/storybook-experimental-vite

konstantinblaesi:
@benmccann when doing 
```
npm install
npm run storybook
```
for that repository, I see traces of webpack. Is storybook somehow ditching the vite builder once this package is installed/enabled/used?

benmccann:
It uses Vite, but is still pulling in webpack anyway. Should hopefully be fixed in Storybook 7

IanVS:
That's right, in Storybook 6, webpack is still used to build the "manager" part of the UI, which no longer happens in SB 7.

IanVS:
@j3rem1e what would you think about adding vite-builder support into this addon?  Currently, we have a [custom vite-plugin](https://github.com/storybookjs/storybook/blob/next/code/frameworks/svelte-vite/src/plugins/csf-plugin.ts) in the `svelte-vite` framework (which I realized recently isn't being used, though it is in 6.5), which imports some files from this addon and duplicates some of what this addon is doing.  

What would you think about making this addon-svelte-csf package a bit more general, including the vite plugin here, and adding a `viteFinal` alongside the current webpack config?  It would mean a few more optional peer dependencies perhaps, but it might be nice to have the related logic all in one place, and one addon to tell people to install.  What do you think?  Should I work on a PR?

j3rem1e:
@IanVS yep, that's a good idea ^^

benmccann:
https://github.com/storybookjs/addon-svelte-csf/pull/72 was merged and said it closes this issue. I'm confused as to why GitHub didn't auto-close this issue. Maybe it was just a GitHub glitch? Should this issue be closed or is there more to do here?

IanVS:
I think the problem is that I used the storybook monorepo convention of `issue: #xyz` instead of `closes: #xyz`, and the same scripts that run there, don't run here.  I do believe this can be closed now, thanks for mentioning it.

PR #72, opened 2022-10-06
Add vite support

IanVS:
Closes https://github.com/storybookjs/addon-svelte-csf/issues/64

This adds the vite plugin from https://github.com/storybookjs/storybook/blob/next/code/frameworks/svelte-vite/src/plugins/csf-plugin.ts into this repo, and sets up a `viteFinal` to load it on vite storybook projects.  There are a few things I'd like to call out here:

1. Without setting a `target` for `@babel/preset-env`, it was trying to add regeneratorRuntime into the built code, which failed because we're not adding it as a polyfill.  I'm imagining that this PR, if merged, would be for use with storybook 7.0+, which supports Chrome 100+ and Node 14+, so those are the targets I added.  This means it's probably a breaking change for this package, so I wanted to call that out explicitly.
2. I needed to turn off some typescript eslint rules, because this repo does not use typescript.
3. If/when we merge this, I'll remove the vite plugin from the storybook vite builder, so that they're not duplicated.  
4. Dual-builder support is a little bit of a bummer in terms of dependencies.  I added magic-string as a dependency because it's used in the plugin, and it doesn't make sense to have as a peer dep.  The others, I made optional peer deps.  It's safe to assume that vite will be available when viteFinal is run, and `@sveltejs/vite-plugin-svelte` should be as well in a svelte vite project, but I added a little safety check around that one.

I tested this by creating an example svelte project, modifying the vite builder to not inject the plugin and copying its `dist` over to the example, and building this addon and copying it's `dist` over as well.  It worked correctly, but it might be nice to consider some tests or example project in this addon, for testing, as well.  I didn't undertake that here, but can if desired.  

<details>
  <summary>📦 Published PR as canary version: <code>2.0.11--canary.72.ae738ba.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@2.0.11--canary.72.ae738ba.0
  # or 
  yarn add @storybook/addon-svelte-csf@2.0.11--canary.72.ae738ba.0
  ```
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,28 +34,28 @@
     "release": "yarn build && auto shipit"
   },
   "dependencies": {
+    "@babel/runtime": "^7.17.5",
     "magic-string": "^0.26.6",
-    "ts-dedent": "^2.0.0",
-    "@babel/runtime": "^7.17.5"
+    "ts-dedent": "^2.0.0"
   },
   "devDependencies": {
     "@auto-it/released": "^10.32.6",
     "@babel/cli": "^7.17.6",
     "@babel/core": "^7.17.5",
+    "@babel/plugin-transform-runtime": "^7.17.5",
     "@babel/preset-env": "^7.16.11",
     "@babel/preset-react": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
-    "@babel/plugin-transform-runtime": "^7.17.5",
     "@storybook/addon-essentials": "^7.0.0-alpha.29",
     "@storybook/addon-interactions": "^7.0.0-alpha.29",
     "@storybook/addon-storyshots": "^7.0.0-alpha.29",
-    "@storybook/core-server": "^7.0.0-alpha.29",
     "@storybook/core-client": "^7.0.0-alpha.29",
+    "@storybook/core-server": "^7.0.0-alpha.29",
     "@storybook/eslint-config-storybook": "^3.1.2",
     "@storybook/jest": "^0.0.10",
-    "@storybook/testing-library": "^0.0.9",
     "@storybook/svelte": "^7.0.0-alpha.29",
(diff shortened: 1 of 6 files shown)


JReinhold (approved):
This looks good to me, I think this is the correct way forward!

I changed the base branch to `next` which caused a few conflicts though.

JReinhold:
@IanVS do you want to take a look at the merge conflicts, or instead target this back to 6.5? To me it looks like the conflicts are mostly about the upgrade to 7.0.

package.json
```diff
@@ -68,20 +68,30 @@
...
     "svelte": "^3.50.0",
     "svelte-jester": "^2.3.2",
     "svelte-loader": "^3.1.2",
-    "typescript": "^3.9.7"
+    "typescript": "^3.9.7",
+    "vite": "^3.1.4"
   },
   "peerDependencies": {
     "@storybook/svelte": ">=6.4.20",
     "@storybook/theming": ">=6.4.20",
```
IanVS, line 80: Is this correct?  Shouldn't these be 7.0?

IanVS:
Thanks for the nudge @JReinhold, I think this is ready for re-review.  I didn't have a chance to test it all out again.

package.json
JReinhold: You're right. Fixing in another PR

IanVS:
@JReinhold @j3rem1e either of you care to give this a last lookover?  Should we try merging and getting things ready on the builder-vite side before the upcoming 7.0 beta release?

shilman:
:rocket: PR was released in [`v3.0.0-next.1`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v3.0.0-next.1) :rocket:

Issue #73, opened 2022-10-17
[Bug] cannot use strict typescript types


RSWilli:
### Describe the bug

When i set `strict: true` in my tsconfig.json I get the following error in every stories file:

```
Argument of type 'typeof Story' is not assignable to parameter of type 'ConstructorOfATypedSvelteComponent'.
  Types of construct signatures are incompatible.
    Type 'new (options: ComponentConstructorOptions<StoryProps>) => Story' is not assignable to type 'new (args: { target: any; props?: any; }) => ATypedSvelteComponent'.
      Construct signature return types 'Story' and 'ATypedSvelteComponent' are incompatible.
        The types of '$on' are incompatible between these types.
          Type '<K extends never>(type: K, callback: (e: {}[K]) => void) => () => void' is not assignable to type '(event: string, handler: ((e: any) => any) | null | undefined) => () => void'.
            Types of parameters 'callback' and 'handler' are incompatible.
              Type '((e: any) => any) | null | undefined' is not assignable to type '(e: never) => void'.
                Type 'undefined' is not assignable to type '(e: never) => void'.

Possible causes:
- You use the instance type of a component where you should use the constructor type
- Type definitions are missing for this Svelte Component. If you are using Svelte 3.31+, use SvelteComponentTyped to add a definition:
  import type { SvelteComponentTyped } from "svelte";
  class ComponentName extends SvelteComponentTyped<{propertyName: string;}> {}ts(2345)
``` 

### Steps to reproduce the behavior

1. add `strict: true` in tsconfig.json

### Expected behavior

no error

### Screenshots and/or logs

If applicable, add screenshots and/or logs to help explain your problem.

### Environment

```
  System:
    OS: Linux 5.15 Arch Linux
    CPU: (8) x64 Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz
    Memory: 6.15 GB / 31.22 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.2.0 - ~/.nvm/versions/node/v16.2.0/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 8.3.0 - ~/.nvm/versions/node/v16.2.0/bin/npm
  Browsers:
    Chromium: 106.0.5249.103
    Firefox: 105.0.3
  npmPackages:
    @storybook/addon-svelte-csf: ^2.0.8 => 2.0.8 
[...]

PR #74, opened 2022-10-18
Make types compatible with TypeScript `strict` mode

RSWilli:
see #73

changed the events type parameter to any for Template and Story, since those are the default, as per [this line](https://github.com/sveltejs/svelte/blob/master/src/runtime/internal/dev.ts#L213)


kasperpeulen (approved):
LGTM

shilman:
:rocket: PR was released in [`v2.0.11`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v2.0.11) :rocket:

Issue #88, opened 2023-01-10
[Bug] typescript not working in stories for svelte


mantismamita:
### Steps to reproduce the behavior

Reproducable in https://github.com/mantismamita/storybook-svelte-bugreproduction/tree/typescript-in-story

1. npm install
2. npm run storybook

I see 
```
WARN 🚨 Extraction error on src/stories/Visual.stories.svelte: Unexpected token (2:14)
WARN 1: <script lang="ts">
WARN 2:   import type { Visual as VisualType } from 'visual';
WARN                  ^
WARN 3: 
WARN 4:   import { Meta, Story, Template } from '@storybook/addon-svelte-csf';
```

### Expected behavior

I expect storybook to load 

### Screenshots and/or logs

```
@storybook/cli v7.0.0-beta.21

info => Starting manager..
info => Serving static files from ././static at /
WARN 🚨 Extraction error on src/stories/Visual.stories.svelte: Unexpected token (2:14)
WARN 1: <script lang="ts">
WARN 2:   import type { Visual as VisualType } from 'visual';
WARN                  ^
WARN 3: 
WARN 4:   import { Meta, Story, Template } from '@storybook/addon-svelte-csf';
CompileError [ParseError]: Unexpected token
    at error (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:13771:20)
    at Parser$1.error (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:13847:7)
    at Parser$1.acorn_error (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:13841:12)
    at Object.read_script [as read] (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:7372:14)
    at tag (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:12891:31)
    at new Parser$1 (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:13807:17)
    at parse$I (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:13934:20)
    at Object.compile (/Users/kirstencassidy/Work/storybook-svelte-bugreproduction/node_modules/svelte/compiler.js:40767:17)
[...]

PR #94, opened 2023-03-08
fix: preprocess svelte file during indexing

ysaskia:
Closes https://github.com/storybookjs/addon-svelte-csf/issues/88

It fixes the issue where svelte files wasn't pre-processed during indexing making it impossible to load `*.stories.svelte` files with `lang="ts"` for example. Indeed this implementation is a very naïve approach. It walking up the directory tree to look for the first `svelte.config.js` file. It should solve 90% of the cases since svelte apps don't work without it. Indeed projects with exotic location or specifying a new svelte config in the storybook main file are not taken into account here. 

Hope it helps

<details>
  <summary>📦 Published PR as canary version: <code>2.0.12--canary.94.71b3707.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@2.0.12--canary.94.71b3707.0
  # or 
  yarn add @storybook/addon-svelte-csf@2.0.12--canary.94.71b3707.0
  ```
</details>

### DIFF
--- a/src/config-loader.js
+++ b/src/config-loader.js
@@ -17,7 +17,7 @@ export async function loadSvelteConfig() {
   const configFile = await findSvelteConfig();
 
   // no need to throw error since we handle projects without config files
-  if (configFile !== undefined) {
+  if (configFile === undefined) {
     return undefined;
   }
 
--- a/src/config-loader.js
+++ b/src/config-loader.js
@@ -1,0 +1,171 @@
+// This file is a rewrite of `@sveltejs/vite-plugin-svelte` without the `Vite`
+// parts: https://github.com/sveltejs/vite-plugin-svelte/blob/e8e52deef93948da735c4ab69c54aced914926cf/packages/vite-plugin-svelte/src/utils/load-svelte-config.ts
+import { logger } from '@storybook/client-logger';
+import { createRequire } from 'module';
+import path from 'path';
+import fs from 'fs';
+import { pathToFileURL, fileURLToPath } from 'url';
+
+/**
+ * Try find svelte config and then load it.
+ *
+ * @returns {import('@sveltejs/kit').Config}
+ * Returns the svelte configuration object.
+ */
+export async function loadSvelteConfig() {
+  const configFile = findSvelteConfig();
+  let err;
+
+  // try to use dynamic import for svelte.config.js first
+  if (configFile.endsWith('.js') || configFile.endsWith('.mjs')) {
+    try {
+      return importSvelteOptions(configFile);
+    } catch (e) {
+      logger.error(`failed to import config ${configFile}`, e);
+      err = e;
+    }
+  }
+  // cjs or error with dynamic import
+  if (configFile.endsWith('.js') || configFile.endsWith('.cjs')) {
+    try {
+      return requireSvelteOptions(configFile);
+    } catch (e) {
+      logger.error(`failed to require config ${configFile}`, e);
+      if (!err) {
+        err = e;
+      }
+    }
+  }
+  // failed to load existing config file
+  throw new Error(`failed to load config ${configFile}`, { cause: err });
+}
+
+const importSvelteOptions = (() => {
+  // hide dynamic import from ts transform to prevent it turning into a require
+  // see https://github.com/microsoft/TypeScript/issues/43329#issuecomment-811606238
+  // also use timestamp query to avoid caching on reload
+  const dynamicImportDefault = new Function(
+    'path',
+    'timestamp',
+    'return import(path + "?t=" + timestamp).then(m => m.default)'
+  );
+
+  /**
+   * Try import specified svelte configuration.
+   *
+   * @param {string} configFile
+   * Absolute path of the svelte config file to import.
+   *
+   * @returns {import('@sveltejs/kit').Config}
+   * Returns the svelte configuration object.
+   */
+  return async (configFile) => {
+    const result = await dynamicImportDefault(
+      pathToFileURL(configFile).href,
+      fs.statSync(configFile).mtimeMs
+    );
+
+    if (result != null) {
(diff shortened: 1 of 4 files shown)


JReinhold:
Thanks for this, this is great! 🙏 

A few initial thoughts before I've tried it out.

This implementation is builder agnostic which is nice. But when looking at [the implementation of `loadSvelteConfig` from `@sveltejs/vite-plugin` ](https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/src/utils/load-svelte-config.ts) there's quite a lot of edge cases that it handles, that this PR doesn't. On one hand it would be nice to be builder agnostic. On the other hand, off-loading all the complexity to the official Vite plugin would make it easier to maintain here, and it would probably behave more predictively.

That would leave out Webpack users though. Is there a compromise where we somehow detect if this is a Vite or Webpack project, and use either the Vite plugin or this "naïve" approach in a Webpack project?

---

At a minimum I think we want to support `svelte.config.mjs` and `svelte.config.cjs` in addition to `svelte.config.js` though.

---

I'm not sure the whole `findUp` solution is necessary. At least in a Vite based setup, `svelte.config.js` always have to be in the CWD for the Svelte app to read it (not Storybook, but the actual app), and so I would think we can keep that same limitation here? I'm not sure if it's the same in a Webpack-based project though.

JReinhold:
I think that sounds like a good plan, looking forward to see your work!

If down the road users complain that the preprocessing doesn't match their Vite config, then we can do another pass at integrating deeper with that, but for now let's wait and see.

src/preset/indexer.js
```diff
@@ -1,8 +1,15 @@
-import { extractStories } from '../parser/extract-stories';
 import fs from 'fs-extra';
+import * as svelte from 'svelte/compiler';
+import { extractStories } from '../parser/extract-stories';
+import { loadSvelteConfig } from '../config-loader';
 
 export async function svelteIndexer(fileName, { makeTitle }) {
   let code = (await fs.readFile(fileName, 'utf-8')).toString();
+  const svelteOptions = loadSvelteConfig();
```
JReinhold, line 8: This resolves a promise, which made it not work for me. When I added `await` here, it worked.

```suggestion
  const svelteOptions = await loadSvelteConfig();
```

src/config-loader.js
```diff
@@ -0,0 +1,171 @@
...
+ */
+function findSvelteConfig() {
+  const lookupDir = process.cwd();
+  let configFiles = getConfigFiles(lookupDir);
+
+  if (configFiles.length === 0) {
+    configFiles = getConfigFilesUp();
+  }
+  if (configFiles.length === 0) {
+    throw new Error('no svelte config found');
```
JReinhold, line 120: Since this error is not caught it means that a Svelte config is now required, even for users that don't want preprocessing. I don't think that's what we want, I guess this should return `undefined` here instead, which should propagate to the indexer. The indexer already handles undefined `svelteOptions`.

src/config-loader.js
```diff
@@ -0,0 +1,171 @@
...
+ * @param {string} lookupDir
+ * Directory in which to look for svelte files.
+ *
+ * @returns {string[]}
+ * Returns an array containing all available config files.
+ */
+function getConfigFiles(lookupDir) {
+  return knownConfigFiles
+    .map((candidate) => path.resolve(lookupDir, candidate))
+    .filter((file) => fs.existsSync(file));
```
JReinhold, line 168: We might gain some perf my making this async, and I don't think it would be a problem since the whole config loading is async anyway. It would make this function a bit more complex, but I think that's worth it.
```suggestion
    .filter((file) => fs.exists(file));
```

src/config-loader.js
```diff
@@ -0,0 +1,171 @@
...
+    configFiles = getConfigFilesUp();
+  }
+  if (configFiles.length === 0) {
+    throw new Error('no svelte config found');
+  } else if (configFiles.length > 1) {
+    logger.warn(
+      `found more than one svelte config file, using ${configFiles[0]}. ` +
+        'you should only have one!',
+      configFiles
+    );
```
JReinhold, line 126: This is just a nitpick on messaging

```suggestion
    logger.warn(
      'Multiple svelte configuration files were found, which is unexpected. The first one will be used.',
      configFiles
    );
```

JReinhold (changes requested):
Tried it out, a few minor improvements needed, but otherwise great work!

ysaskia:
Hey @JReinhold! Thanks a lot for the review. It is now updated accordingly. I used `fs-extra.pathExists` instead of native `fs.exists` since is deprecated and `fs-extra` already used in the package. Only one remark `fs-extra` is not in the package dependencies. Is this intentional?

src/config-loader.js
```diff
@@ -0,0 +1,188 @@
...
+ * Try find svelte config and then load it.
+ *
+ * @returns {import('@sveltejs/kit').Config | undefined}
+ * Returns the svelte configuration object.
+ */
+export async function loadSvelteConfig() {
+  const configFile = await findSvelteConfig();
+
+  // no need to throw error since we handle projects without config files
+  if (configFile !== undefined) {
```
JReinhold, line 20:
```suggestion
  if (configFile === undefined) {
```

A pretty critical typo 😅

JReinhold (changes requested):
Looks good! If you can apply my suggestion and add `fs-extra` as a dependency as it should be, this should be ready to go!

shilman:
:rocket: PR was released in [`v3.0.0-next.4`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v3.0.0-next.4) :rocket:

JReinhold:
Thanks for your hard work @ysaskia ❤️

PR #113, opened 2023-07-19
Fix generated ESM, revamp build system

### LINKED ISSUE #111: Module errors when running in storybook 7.1
### Describe the bug

This may be a mite early, but this add-on errors with module issues on storybook@7.1.0-rc.0. See logs below.

Couple things looked odd to me: Seems a little weird that the `cjs` build of the add-on is using `import` statements. I also notice the package.json doesn't declare any `exports`.

1. Repo: https://github.com/hobbes7878/addon-svelte-csf-repro/tree/main
2. Install deps and run `npm run dev`
3. See error immediately

### Expected behavior

Should not throw module errors.

### Screenshots and/or logs

```
ERR! /Users/jon/Scripts/addon-svelte-csf-repro/node_modules/@storybook/addon-svelte-csf/dist/cjs/preset/index.js:1
ERR! import { svelteIndexer } from './indexer.js';
ERR! ^^^^^^
ERR! 
ERR! SyntaxError: Cannot use import statement outside a module
ERR!     at Object.compileFunction (node:vm:360:18)
ERR!     at wrapSafe (node:internal/modules/cjs/loader:1094:15)
ERR!     at Module._compile (node:internal/modules/cjs/loader:1129:27)
ERR!     at Module._extensions..js (node:internal/modules/cjs/loader:1219:10)
ERR!     at Object.newLoader (/Users/jon/Scripts/addon-svelte-csf-repro/node_modules/esbuild-register/dist/node.js:2262:9)
ERR!     at Object.extensions..js (/Users/jon/Scripts/addon-svelte-csf-repro/node_modules/esbuild-register/dist/node.js:4807:24)
ERR!     at Module.load (node:internal/modules/cjs/loader:1043:32)
ERR!     at Function.Module._load (node:internal/modules/cjs/loader:878:12)
ERR!     at Module.require (node:internal/modules/cjs/loader:1067:19)
ERR!     at require (node:internal/modules/cjs/helpers:103:18)
```

### Environment

 - OS: MacOS
 - Node.js version: 16.19.0
 - NPM version: 8.19.3
 - Browser (if applicable): n/a
 - Browser version (if applicable): n/a
 - Device (if applicable): n/a

hobbes7878:
Attempt to close #111. 

Done:
- [x] Updates build system using [@sveltejs/package](https://www.npmjs.com/package/@sveltejs/package).
- [x] Makes package an ES module.
- [x] Converts all the scripts to typescript and fixes any outstanding type issues.
- [x] Updates a buncha dev dependencies. 
- [x] Working with storybook 7.1, updated to use Vite framework.
- [x] Fix/upgrade tests (though, drops snapshot tests in favour of using Chromatic)

<details>
  <summary>📦 Published PR as canary version: <code>3.0.5--canary.113.4269195.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@3.0.5--canary.113.4269195.0
  # or 
  yarn add @storybook/addon-svelte-csf@3.0.5--canary.113.4269195.0
  ```
</details>


JReinhold:
This is looking great @hobbes7878 ! 🥰

It's going to be interesting to see if this will still work in a Webpack-based project, but if not, it would be okay to drop support for that, we're seeing very little usage of Storybook 7 + Webpack + Svelte + this addon.

hobbes7878:
Thanks @JReinhold. Happy if it helps.

On tests, moving the whole package to ECMAScript modules has made it a little tricky to work with Jest. Right now, I've replaced it with vitest for all the unit tests, which was easier to get going. 

That leaves the snapshot tests. I wonder if Chromatic is a good enough replacement for the storyshots. I've updated this branch with that in mind for now, but happy if someone wants to give me any hints/examples how to bring those back...

hobbes7878:
> It's going to be interesting to see if this will still work in a Webpack-based project, but if not, it would be okay to drop support for that, we're seeing very little usage of Storybook 7 + Webpack + Svelte + this addon.

Yeah, I'm not sure, but there weren't any webpack-specific plugins in it originally, so as long as webpack can tolerate the move to ESM and this did _previously_ work there, it _should_ be fine??

davemackintosh:
I've just installed this and I can confirm the issue is fixed for me. Thank you! 🙏

davemackintosh:
I've just installed this and I can confirm the issue is fixed for me. Thank you! 🙏

olafurw:
Tested this also on a branch that had this issue. Fixes the issue for me.

package.json
```diff
@@ -11,9 +11,21 @@
     "url": "https://github.com/storybookjs/addon-svelte-csf"
   },
   "license": "MIT",
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
-  "browser": "dist/cjs/index.js",
```
ndelangen, line 16: You say this works in storybook?
I would expect this to be broken, but honestly I've not yet tried, so maybe I'm wrong.

What i know of the preset-loading mechanism in storybook is what when an addon-entry is added, storybook will try resolving `${name}/preset` (with node).
It will do the same for `manager` and `preview` FYI.

Maybe with `type: module` these exports fields are actually respected/used?

ndelangen (approved):

ilsubyeega (changes requested):

package.json
hobbes7878: Yeah, `type: "module"` is the key there as far as I understand it, which is mostly from [sindresorhus's explanation](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#faq)...

But yes, working in Node 16+ with Storybook 7.1+ in the repo.

src/plugins/vite-svelte-csf.ts
```diff
@@ -7,7 +7,7 @@ import { fileURLToPath } from 'url';
 import { getNameFromFilename } from '../parser/svelte-stories-loader.js';
 import { readFileSync } from 'fs';
 
-const parser = fileURLToPath(new URL('../parser/collect-stories.js', import.meta.url));
+const parser = fileURLToPath(new URL('../parser/collect-stories.js', import.meta.url)).replace(/\\/g, "\\\\");
```
ndelangen, line 10: I'd recommend making this very obvious why this is the way it is perhaps with a clear code-comment?, and possibly even ensure this continues working with a unit-test.

hobbes7878: Added handling across any dynamic imports in https://github.com/storybookjs/addon-svelte-csf/pull/113/commits/fa5505bda90c1c1d06a6a6b3d508b82d6813ef14. 

@ilsubyeega don't know if you'd be able to help with a test?

ilsubyeega: I can say the current branch works on my machine (win11-x64) with `yarn test` and storybook live view.
Also, I'm not sure about creating other test tasks for this.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "@tsconfig/svelte/tsconfig.json",
+  "compilerOptions": {
```
JReinhold, line 3: Missing `strict: true` maybe? Or does that not work for some reason? I see it's explicitly `false` in [`@tsconfig/svelte`](https://www.npmjs.com/package/@tsconfig/svelte)

JReinhold (changes requested):
Absolutely fantastic job, I also love the way you've cleaned up the code all over the place. The "conversion" from storyshots to Chromatic is also \*chef's kiss\* (and Jest -> vitest), storyshots really shouldn't be used anymore.

I have left some comments in the code, and then a big issue is that the Storybook fails to build, which is evident by [the Chromatic check on this PR failing](https://github.com/storybookjs/addon-svelte-csf/actions/runs/5668946724/job/15360661875), and I was able to reproduce it locally. It's failing with some CJS->ESM shenanigans in `cli-table3`. I was able to fix it locally by upgrading all Storybook dependencies from `7.1.0` to `7.1.1`, could you try that? I'll investigate on the core end why this is.

PS. @hobbes7878 would you like to become a maintainer of this project?

tsconfig.json
hobbes7878: Looks like `strict: true` is fine: https://github.com/sveltejs/kit/issues/4395

Will add.

package.json
```diff
@@ -22,68 +34,65 @@
...
-    "@storybook/svelte": "^7.0.0-beta.0 || ^7.0.0-rc.0 || ^7.0.0",
-    "@storybook/theming": "^7.0.0-beta.0 || ^7.0.0-rc.0 || ^7.0.0",
-    "@sveltejs/vite-plugin-svelte": "^1.0.0 || ^2.0.0",
+    "@storybook/svelte": "^7.0.0",
+    "@storybook/theming": "^7.0.0",
+    "@sveltejs/vite-plugin-svelte": "^2.0.0",
     "svelte": "^3.50.0 || ^4.0.0",
     "svelte-loader": "^3.1.2",
-    "vite": "^3.0.0 || ^4.0.0"
+    "vite": "^4.0.0"
```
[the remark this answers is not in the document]
hobbes7878: Widened that range back.

package.json
```diff
@@ -22,68 +34,65 @@
...
+    "vite": "^4.4.3",
+    "vitest": "^0.33.0"
   },
   "peerDependencies": {
-    "@storybook/svelte": "^7.0.0-beta.0 || ^7.0.0-rc.0 || ^7.0.0",
-    "@storybook/theming": "^7.0.0-beta.0 || ^7.0.0-rc.0 || ^7.0.0",
-    "@sveltejs/vite-plugin-svelte": "^1.0.0 || ^2.0.0",
+    "@storybook/svelte": "^7.0.0",
+    "@storybook/theming": "^7.0.0",
+    "@sveltejs/vite-plugin-svelte": "^2.0.0",
```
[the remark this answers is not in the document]
hobbes7878: Widened that range back.

hobbes7878:
Thanks @JReinhold. Addressed all those suggestions, I think.

I'm still seeing that `cli-table3` issue with 7.1.1 Storybook [in the Chromatic workflow](https://github.com/storybookjs/addon-svelte-csf/actions/runs/5700009224/job/15449455253?pr=113#step:4:56). Storybook is building fine for me locally. Will keep investigating...

Re: maintainer, I don't mind, sure!

hobbes7878:
> I have left some comments in the code, and then a big issue is that the Storybook fails to build, which is evident by [the Chromatic check on this PR failing](https://github.com/storybookjs/addon-svelte-csf/actions/runs/5668946724/job/15360661875), and I was able to reproduce it locally. It's failing with some CJS->ESM shenanigans in `cli-table3`. I was able to fix it locally by upgrading all Storybook dependencies from `7.1.0` to `7.1.1`, could you try that? I'll investigate on the core end why this is.

Looks like this isn't fixed just by updating storybook to `7.1.1`. But there's something really weird going on with yarn's dependency resolution here for `string-width`.

I can consistently reproduce this issue based on the existence of `yarn.lock`.

So if `yarn.lock` exists and you clear out `node_modules/` and re-`yarn` to install fresh deps from the lockfile, you get the `ERR_REQUIRE_ESM` error in `cli-table3` requiring `string-width`. That's what's of course happening in the workflow, so that busts out.

If I delete `yarn.lock` (or change a dep so it's regenerated) and then `yarn`, no error. There is no difference in the lockfiles in any of these scenarios.

JReinhold:
Great work!

That sounds super strange, I have no idea why it behaves that way. Could this just be a bug in Yarn 1? I mean, it hasn't been updated for a while.

Would you be open to migrating to pnpm here and see if that automatically fixes the issue? I don't think there's any deeper integrations with `yarn` at play here, so it should be a simple replacement, but if it's not I'm okay with the current solution you have for now.

hobbes7878:
Not gonna lie, it took a couple tries but migrated to pnpm, which is running fine without any dep overloads. 👍

.github/workflows/release.yml
```diff
@@ -7,22 +7,25 @@ jobs:
...
+          version: 8
+          run_install: true
 
       - name: Create Release
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: |
-          yarn release
+          npm run release
```
JReinhold, line 31: shouldn't all these `npm run` be `pnpm run` instead to use the correct package resolution, or am I misunderstanding something? Same for the lines in `package.json`?

hobbes7878: Yeah, that was silly of me. Done in [4269195](https://github.com/storybookjs/addon-svelte-csf/pull/113/commits/4269195b43956bc2f7401f480a9012c9ed660581).

JReinhold (approved):

JReinhold:
Thanks for your hard work and patience @hobbes7878!

oscard0m:
It works! Thanks for the amazing job on this @hobbes7878 🌟 
And thanks to the rest of participants as well 
🎖️ @JReinhold 
🎖️ @ilsubyeega
🎖️ @ndelangen
🎖️ @shilman
🎖️ @olafurw
🎖️ @davemackintosh

PR #115, opened 2023-08-01
fix: swap order of types to avoid module not found

paoloricciuti:
Right now the version 3.0.5 gives you errors in VSCode because the types are exported last in the main export of `package.json`.

Checking the package on publint you can see that this will break typings for TS.

https://publint.dev/@storybook/addon-svelte-csf@3.0.5

I checked locally that swapping the order fixes the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
   "type": "module",
   "exports": {
     ".": {
-      "svelte": "./dist/index.js",
-      "types": "./dist/index.d.ts"
+      "types": "./dist/index.d.ts",
+      "svelte": "./dist/index.js"
     },
     "./preset": {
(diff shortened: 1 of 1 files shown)


hobbes7878:
Yeah, this is right. Thanks for spotting.

@JReinhold, worth another patch bump, this, I think.

shilman:
:rocket: PR was released in [`v3.0.6`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v3.0.6) :rocket:

paoloricciuti:
> 🚀 PR was released in [`v3.0.6`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v3.0.6) 🚀

Thanks for the quick release!  🙏🏻

Issue #121, opened 2023-08-11
[Bug] Type error `typeof Meta` when used with Svelte 3


mcmxcdev:
### Describe the bug

I am encountering TS issues with all imports from `@storybook/addon-svelte-csf`:

```
Argument of type 'typeof Meta' is not assignable to parameter of type 'ConstructorOfATypedSvelteComponent'.
  Type 'typeof Meta' provides no match for the signature 'new (args: { target: any; props?: any; }): ATypedSvelteComponent'.

Possible causes:
- You use the instance type of a component where you should use the constructor type
- Type definitions are missing for this Svelte Component. If you are using Svelte 3.31+, use SvelteComponentTyped to add a definition:
  import type { SvelteComponentTyped } from "svelte";
  class ComponentName extends SvelteComponentTyped<{propertyName: string;}> {}
```

The exact same message with a different `typeof` is repeated for `Template` and `Story` imports.

### Expected behavior

No TS issues

### Environment

 - OS: Ubuntu 23.04
 - Node.js version: v18.17.1
 - NPM version: pnpm 8.6.10
 - Browser (if applicable): Brave
 - Browser version (if applicable): Version 1.56.20 Chromium: 115.0.5790.171 (Official Build) (64-bit)
 - Device (if applicable): Desktop

### Additional context

Story looks like this:

```svelte
<script lang="ts">
  import { Meta, Template, Story } from '@storybook/addon-svelte-csf'
  import Space from '$lib/components/ui/Space.svelte'
</script>

<Meta title="Actions/Space" component={Space} />

<Template let:args>
  <Space {...args} />
</Template>

<Story name="Horizontal" />

<Story name="Vertical" />
```

Config is as follows:

```ts
import type { StorybookConfig } from '@storybook/sveltekit'

const config: StorybookConfig = {
  stories: [
    {
      directory: '../src/stories/**',
      files: '*.stories.@(ts|svelte)',
      titlePrefix: 'Design System',
    },
  ],
  addons: [
    '@storybook/addon-interactions',
    '@storybook/addon-links',
    '@storybook/addon-a11y',
    '@storybook/addon-svelte-csf',
    {
      name: '@storybook/addon-essentials',
      options: {
        backgrounds: true,
      },
    },
    {
      name: '@storybook/addon-styling',
      options: {},
    },
  ],
  framework: {
    name: '@storybook/sveltekit',
    options: {},
  },
  docs: {
    autodocs: false,
  },
[...]

PR #112, opened 2023-07-14
Update dependencies to support svelte@4

RSWilli:
Closes #109

* Updated the types according to https://github.com/sveltejs/svelte/pull/8512
* Updated the storybook dependencies to support svelte 4, see https://github.com/storybookjs/storybook/releases/tag/v7.0.27
* Use svelte 4 internally, but allow svelte 3 as peer dependency
* use Node 20 in the build pipeline
* I had to remove the webpackFinal stuff in the storybook main.js for it to work and according to [this comment](https://github.com/storybookjs/addon-svelte-csf/issues/109#issuecomment-1623274565) it was not needed either

The included storybook run fine for me locally.

I am not familiar with the build process with github actions, so feel free to edit anything.

PR #127, opened 2023-08-24
Fix: `typeof Meta` in Svelte v3 (Pin Svelte peer dependency to v3)

JReinhold:
Works on https://github.com/storybookjs/addon-svelte-csf/issues/121#issuecomment-1686214199

cc @RSWilli

This PR pins the allowed Svelte version to v3, because it's not possible to provide correct typings for Svelte v3 and v4 at the same time. It also reverts the type changes in #112 so it matches that of Svelte v3.

This will be released as a patch version, and then shortly after #128 that pins to Svelte v4 will be released as a major, requiring any Svelte v4 users to upgrade this addon with the breaking change.

I've tested that this makes types work by installing this canary in a Svelte v3 project and see that it works, where the current latest release of this addon makes the types fail similar to #121.

<details>
  <summary>📦 Published PR as canary version: <code>3.0.10--canary.127.893640b.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@3.0.10--canary.127.893640b.0
  # or 
  yarn add @storybook/addon-svelte-csf@3.0.10--canary.127.893640b.0
  ```
</details>


JReinhold:
@mcmxcdev could you confirm that the canary version fixes your issue?

shilman:
:rocket: PR was released in [`v3.0.10`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v3.0.10) :rocket:

PR #133, opened 2023-08-30
Fix svelte-stories-loader in Windows

### LINKED ISSUE #131: [Bug] Doesn't work on Windows anymore
### Describe the bug

The v3 of this plugin doesn't work on Windows anymore.

- It fails because it can't find "collect-stories.js" with a wrong path
- It fails with stories containing sources with \n

1. Create a storybook with this addon
2. Compile it under windows

### Screenshots and/or logs

```
ModuleNotFoundError: 
Module not found: 
Error: Can't resolve 
'C:\Users\MyApp\projode_modules\@storybook\addon-svelte-csf\dist\parser\collect-stories.js' 
```

NB: "projode_modules" should be "proj\node_module" but \n has been removed

### Environment

 - OS: Windows 
 - Node.js version: v16
 - NPM version: 8.11

### Additional context

It caused by this commit : https://github.com/storybookjs/addon-svelte-csf/commit/0f5bf9878bf80a09efd84276d8157a729ec7bad3
"Fixed" by this one : https://github.com/storybookjs/addon-svelte-csf/commit/fa5505bda90c1c1d06a6a6b3d508b82d6813ef14#diff-c5f119d3c0fad64afb4a542f1fb2bdf0e475cc6191a63712dbfba50bb5af6c3c

However, it's better to replace backslash with juste slash for windows.

After fixing this issue, the stories containing sources with multilines doesn't compile either. The generated stories contains invalid json.
For example:

```javascript
   const { default: parser } = require('C:/Users/MyApp/proj/node_modules/@storybook/addon-svelte-csf/dist/parser/collect-stories.js');
   const __storiesMetaData = parser(Spacer_stories, {"meta":{"title":"Component/Core/Spacer"},"stories":{"spacer":{"storyId":"component-core-spacer--spacer","name":"Spacer","template":false,"source":"<Surface row>\r
\t<Button>A</Button>\r
\t<Spacer/>\r
\t<Button>B</Button>\r
</Surface>","hasArgs":false}},"allocatedIds":["default","Meta","Story","Surface","Button","Spacer"]});
   export default __storiesMetaData.meta;
   export const spacer = __storiesMetaData.stories["spacer"];
```

j3rem1e:
Fix #131


src/parser/svelte-stories-loader.ts
```diff
@@ -4,7 +4,7 @@ import { fileURLToPath } from 'url';
 import { readFileSync } from 'fs';
 
 const parser = fileURLToPath(new URL('./collect-stories.js', import.meta.url))
-  .replace(/\\/g, "\\\\"); // For Windows paths;
+  .replace(/\\/g, '/'); // For Windows paths;
```
benmccann, line 7: maybe call `path.posix` instead

src/parser/svelte-stories-loader.ts
```diff
@@ -59,7 +59,7 @@ function transformSvelteStories(code: string) {
 
   // throws dedent expression is not callable.
   // @ts-ignore
-  return dedent`${codeWithoutDefaultExport}
+  return `${codeWithoutDefaultExport}
```
benmccann, line 62: why remove dedent? if we remove it I think you'll need to remove the comments on the lines above this as well as the import

j3rem1e: There is a regression with dedent from Storybook v7. It doesn't handle correctly quoted backslash. I have added an example in the referenced issue.

Note that the vite loader doesn't use dedent either.

benmccann: will https://github.com/storybookjs/addon-svelte-csf/blob/c68b02cb9279b9ea8dba67eb8dbd3a1cb615896a/src/parser/extract-stories.ts#L131 fail on windows as well?

j3rem1e: Probably. However here where are not generating invalid javacsript. at worst the source will have invalid breaking line, but we really need dedent here.

In fact, this issue is not related to windows and is caused by this commit: https://github.com/storybookjs/addon-svelte-csf/commit/02ce02a7b2e3be5b0b29ef844dc05d38004dbfc4

j3rem1e:
Note: All my instances of Storybook work with this patch, SB7.4, Webpack5 and Svelte 4.2

src/parser/svelte-stories-loader.ts
benmccann: I wonder if it would help to upgrade this project from decent 1.2 to the latest version (1.5.1 if I recall correctly)

j3rem1e: If you're ok with this, let's create a new issue to improve the rendering of the source snippet.

I'd like to close this PR, it's the last issue for upgrading my stack to Svelte 4 & Storybook 7.4 ^^

shilman:
:rocket: PR was released in [`v4.0.1`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.0.1) :rocket:

Issue #129, opened 2023-08-29
[Feature Request] Support autotitle


JReinhold:
Storybook has a feature called autotitle, that sets the title of the component (meta) based on the file path, if no title is explicitly set. It would be nice if this addon supported that feature.

### Implementation

The indexer gets passed a `makeTitle` function in the indexer options (second parameter) to the `indexer` function. That function can be used to generate this title automatically, also taking into account any user title specified.

### Additional context

Implementing this with `makeTitle` would probably also solve the bug at #130 at the same time.

benmccann:
https://github.com/storybookjs/addon-svelte-csf/issues/101 sounds potentially related as well

j3rem1e:
This API didn't exist when I have implemented the indexer. I can have a look at it.

j3rem1e:
I have fixed autotitle here: https://github.com/storybookjs/addon-svelte-csf/pull/134

Issue #101, opened 2023-04-13
[Bug] Empty title doesn’t work


enyo:
### Describe the bug

The storybook docs state that the title doesn’t need to be specified. When omitting the title with this addon however, the story stops working.

### Steps to reproduce the behavior

Create a story and don’t specify the title in `<Meta>`

### Expected behavior

The title should be auto generated.

j3rem1e:
It should be fixed here #134

PR #136, opened 2023-09-02
Fix autodocs

### LINKED ISSUE #130: [Bug] `titlePrefix` in advanced story specifiers causes the story to crash with "Didn't find 'xyz' in CSF file"
### Describe the bug

Originally discussed at #120.

If a stories specifier uses [the advanced pattern with a title prefix](https://storybook.js.org/docs/react/configure/sidebar-and-urls#auto-title-prefixes) like this:

```
const config = {
  stories: [
    {
      directory: '../src',
      titlePrefix: 'Custom', // 👈 Configure the title prefix
    },
  ],
};

```

The story will fail to render with "Didn't find 'some-story-id' in CSF file", if it's been written using Svelte CSF.

This happens because the indexer in this addon ignores that title prefix when generating story IDs, but the Storybook CSF processing in the browser does _not_ ignore that title prefix, so the ID it is looking for will not match the IDs in the index.

Reproduction: https://github.com/mcmxcdev/storybookjs-addon-svelte-csf-120

### Screenshots and/or logs

image

### Additional context

This bug can probably be fixed by using the `makeTitle` function that is passed in the indexer options (second parameter) to the indexer function. That function can be used to generate titles automatically, also taking into account any user title specified.

Using `makeTitle` to fix this bug will likely also solve #129 .

j3rem1e:
Fixes #130 #129 #101

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -2,9 +2,11 @@ import type { StorybookConfig } from '@storybook/svelte-vite';
 
 const config: StorybookConfig = {
   framework: '@storybook/svelte-vite',
-  stories: [
-    '../stories/**/*.stories.svelte',
-  ],
+  stories: [{
+    directory: '../stories',
+    files:'**/*.stories.svelte',
+    titlePrefix:'Demo'
+  }],
   addons: [
     '../dist/preset/index.js',
     '@storybook/addon-essentials',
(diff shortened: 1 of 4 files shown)


shilman:
:rocket: PR was released in [`v4.0.2`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.0.2) :rocket:

PR #138, opened 2023-09-12
fix: move MetaProps to its own declaration to allow for overrides

paoloricciuti:
Right now the Meta class doesn't have a Props interface. This make it impossible to add overrides to the props (i'm using interface merging to provide intellisense for some parameters)

With this simple change the users will be able to augment the the typing of the Meta component

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -44,7 +44,7 @@ interface TemplateProps extends BaseAnnotations<any, DecoratorReturnType> {
     id?: string;
 }
 
-interface MetaProps extends BaseMeta<any> extends BaseAnnotations<any, DecoratorReturnType> { }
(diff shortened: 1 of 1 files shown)


oscard0m (approved):

oscard0m:
✅ but wait for @JReinhold's review

paoloricciuti:
> FYI the Meta tag will be deprecated soon.

Oh yeah I saw that discussion...well until then 😄 also this could help to keep the typing this way when the meta is moved to the module script

JReinhold (approved):
LGTM!

paoloricciuti:
Do I need to do something else before merging?

shilman:
:rocket: PR was released in [`v4.0.4`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.0.4) :rocket:

PR #139, opened 2023-09-13
fix: MetaProps typing

paoloricciuti:
I'm sorry, but i just realized that while #138 works fine the correct way of extending two classes should be this one.

Sorry again for the double PR 🙏🏼

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -44,7 +44,7 @@ interface TemplateProps extends BaseAnnotations<any, DecoratorReturnType> {
     id?: string;
 }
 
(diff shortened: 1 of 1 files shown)


oscard0m (approved):
LGTM, but wait for @JReinhold's review.

In terms of releases I'm not sure how this should be managed.

JReinhold (approved):
Thanks for the quick follow-up 💪

shilman:
:rocket: PR was released in [`v4.0.5`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.0.5) :rocket:

PR #153, opened 2023-11-02
Migrate to new story indexer API

### LINKED ISSUE #147: Migrate to new story indexer API
We've been working on a new story indexer API for some time, that is more official and public and actually documented. The API currently used by this addon was "internal" and undocumented. It has officially been deprecated in Storybook 7.5.0 in favor of the new API.

The new API is documented here: https://storybook.js.org/docs/svelte/api/main-config-indexers

I suggest we migrate this addon to the new API for multiple reasons:

1. Remove the deprecation warning for users on Storybook >7.5.0
1. Be ready for Storybook 8.
1. Provide feedback to the API. We might uncover different usage patterns here that would warrant a change in the API.

The new API is not very different from the old one so migrating should hopefully be straight forward. The biggest difference is that an indexer now doesn't return an instance of `CSFFile` (in this case `{ meta: ..., stories: ...}`) but rather a list of more detailed `IndexInput`, allowing the indexer to specify some properties that was not exposed before.
The documentation also details how transpiling CSF files is necessary etc. but that should already have been taken care of in this addon as that remains the same.

The deprecation warning is implemented in way that allows addons to support both the new and old API simultaneously if they want to. Storybook 7.4.0+ will use the `experimental_indexers` if found and fallback to `storyIndexers` with a warning. Older versions will use `storyIndexers` with no warnings.

The plan is to completely remove support for the old API in Storybook 8.0.

This API is behind an experimental flag right now until we feel like the API is stable It's not because it could potentially be removed since Storybook Core uses it for all story indexing internally.

j3rem1e:
Closes #147


src/preset/indexer.ts
```diff
@@ -4,24 +4,61 @@ import { extractStories } from '../parser/extract-stories.js';
 import fs from 'fs-extra';
 import { loadSvelteConfig } from '../config-loader.js';
 import { storyNameFromExport, toId } from '@storybook/csf';
+import { IndexInput, IndexedCSFFile, IndexerOptions } from '@storybook/types';
 
-export async function svelteIndexer(fileName, { makeTitle }) {
+/**
+ *
+ */
```
benmccann, line 11:
```suggestion
```

src/preset/index.ts
```diff
@@ -65,3 +69,17 @@ export const storyIndexers = async (indexers) => {
     ...(indexers || []),
   ];
 };
+
+/**
+ * Storybook >= 7.4
+ */
+export const experimental_indexers = (indexers) => {
```
benmccann, line 76: should this use camelCase to match the others?

j3rem1e: I agree, but it's the name expected by Storybook: https://github.com/storybookjs/storybook/blob/next/code/lib/core-server/src/utils/getStoryIndexGenerator.ts#L28

benmccann (approved):

shilman:
:rocket: PR was released in [`v4.0.10`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.0.10) :rocket:

shilman:
:rocket: PR was released in [`v4.0.10`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.0.10) :rocket:

shilman:
:rocket: PR was released in [`v4.0.10`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.0.10) :rocket:

src/preset/index.ts
```diff
@@ -1,5 +1,6 @@
 import { fileURLToPath } from 'url';
-import { svelteIndexer } from './indexer.js';
+import { svelteIndexer, createIndex } from './indexer.js';
+import { create } from '@storybook/theming';
```
JReinhold, line 3: This import was probably a mistake?

JReinhold (approved):
Niiice!! 👌

PR #159, opened 2023-11-23
Update versions of peer dependencies to allow latest Vite and Vite Svelte plugin

joekrump:
Fixes https://github.com/storybookjs/addon-svelte-csf/issues/161

Now that [Vite 5 has been released](https://vitejs.dev/blog/announcing-vite5), it would be nice to have te versions of the peer dependencies for it and the Svelte Vite plugin which uses Vite, updated to the latest versions. If this could be published as a new release that would most awesome!

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "@sveltejs/vite-plugin-svelte": "^3.0.0",
     "svelte": "^4.0.0",
     "svelte-loader": "^3.1.2",
-    "vite": "^5.0.0"
+    "vite": "^4.0.0 || ^5.0.0"
   },
   "peerDependenciesMeta": {
     "@sveltejs/vite-plugin-svelte": {
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
   "peerDependencies": {
     "@storybook/svelte": "^7.0.0",
     "@storybook/theming": "^7.0.0",
-    "@sveltejs/vite-plugin-svelte": "^3.0.0",
+    "@sveltejs/vite-plugin-svelte": "^2.0.0 || ^3.0.0",
     "svelte": "^4.0.0",
     "svelte-loader": "^3.1.2",
     "vite": "^4.0.0 || ^5.0.0"
--- a/package.json
+++ b/package.json
@@ -100,10 +100,10 @@
   "peerDependencies": {
     "@storybook/svelte": "^7.0.0",
     "@storybook/theming": "^7.0.0",
-    "@sveltejs/vite-plugin-svelte": "^2.0.0",
+    "@sveltejs/vite-plugin-svelte": "^3.0.0",
     "svelte": "^4.0.0",
     "svelte-loader": "^3.1.2",
-    "vite": "^4.0.0"
+    "vite": "^5.0.0"
   },
   "peerDependenciesMeta": {
     "@sveltejs/vite-plugin-svelte": {


j3rem1e:
I don't think it's a good idea to drop support for vite4 now

joekrump:
> I don't think it's a good idea to drop support for vite4 now

@j3rem1e are you able to say a little more about the thinking behind that?

I agree that vite 4 should still be supported since it is widely used. To do this and support vite 5, a new major release of this package could be made that supports vite >=5 while still doing minor release bumps for the current major version that requires vite <= 4. Curious to hear your thoughts on that path forward.

package.json
```diff
@@ -100,10 +100,10 @@
   "peerDependencies": {
     "@storybook/svelte": "^7.0.0",
     "@storybook/theming": "^7.0.0",
-    "@sveltejs/vite-plugin-svelte": "^2.0.0",
+    "@sveltejs/vite-plugin-svelte": "^3.0.0",
     "svelte": "^4.0.0",
     "svelte-loader": "^3.1.2",
-    "vite": "^4.0.0"
+    "vite": "^5.0.0"
```
ai, line 106:
```suggestion
    "vite": "^4.0.0 || ^5.0.0"
```

joekrump: So my only concern here is that this could result in some things breaking for some users. I'm not 💯 sure about this codebase and whether any of the breaking changes apply though, so that may not be the case.

https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md#breaking-changes

matths:
Any news here? It's two weeks and I would appreciate this addon working with Vite5.

package.json
```diff
@@ -100,10 +100,10 @@
   "peerDependencies": {
     "@storybook/svelte": "^7.0.0",
     "@storybook/theming": "^7.0.0",
-    "@sveltejs/vite-plugin-svelte": "^2.0.0",
+    "@sveltejs/vite-plugin-svelte": "^3.0.0",
```
JReinhold, line 103: If no other changes are needed it looks like we can safely support both v2 and v3?
Only having v3 here won't work, as that only supports Vite 5, so even if our Vite peer deps allows 4+5, in the end the package manager will resolve it to only support Vite 5.
```suggestion
    "@sveltejs/vite-plugin-svelte": "^2.0.0 || ^3.0.0",
```

JReinhold (commented):
Looks like we can do this without a breaking change, supporting both versions?
Otherwise I'd expect some logic changes as well somewhere.

I would advice against dropping support for Vite 4 given that SvelteKit still uses it.

JReinhold:
If someone is willing to test this addon in a Vite 5 project and can confirm it works fine, I'd be happy to merge and release this.

matths:
> If someone is willing to test this addon in a Vite 5 project and can confirm it works fine, I'd be happy to merge and release this.

We would need a package to include to test it with our Svelte project. Any advice how to come up with that?

Masstronaut:
@matths could you use the existing release and update the package.json as is done in this PR? It's a 2 line change.

Masstronaut:
I tested this in a new sveltekit project per the sveltekit docs:

```bash
npm create svelte@latest my-app
cd my-app
npm install
```

and in my `package.json` `"devDependencies"` have:

```json
"@sveltejs/kit": "^2.0.0",
"@sveltejs/vite-plugin-svelte": "^3.0.0",
"vite": "^5.0.3",
"svelte": "^4.2.7",
...
```

I then ran `npx storybook init` and `npm i -D @storybook/addon-svelte-csf --force` and can confirm that *storybook seemed to be working fine with no issues*. Will report back when I have a chance to test it on a larger project.

shilman:
:rocket: PR was released in [`v4.1.0`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.1.0) :rocket:

PR #173, opened 2024-02-20
Update README.md

LuisEGR:
Updated the Getting Started instruction number 3, this was quite confusing, initially says add "*.stories.svelte" which if taked literary, one can think of adding this string as a new element to the array, but after closing inspection (because it is not working), one can realize in the example the correct configuration, all this can be avoided if the instructions are clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Actions are automatically registered by Storybook. To be used by this addon, you
 
 1. `npm install --save-dev @storybook/addon-svelte-csf` or `yarn add --dev @storybook/addon-svelte-csf`
 2. In `.storybook/main.js`, add `@storybook/addon-svelte-csf` to the addons array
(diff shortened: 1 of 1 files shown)


JReinhold:
I see what you mean, but I think this proposal could be equally confusing. What about something like

> 3. In `.storybook/main.js`, include `.stories.svelte` files in your `stories` patterns, eg. by changing the patterns to `'../src/**/*.stories.@(js|jsx|ts|tsx|svelte)'`

LuisEGR:
I agree with you, I've just pushed another commit with that change.

shilman:
:rocket: PR was released in [`v4.1.2`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.1.2) :rocket:

PR #179, opened 2024-05-12
chore: Remove `fs-extra` in favor of `node:fs`

xeho91:
I couldn't resist, so I experimented a lot with this addon in order to support Svelte `v5`.
I've refactored the parser locally, to fix the issues I've encountered.

In order to avoid the "big" PR... I'd like to split into smaller PR's, so there's no overhead in the future code review 😅

## What else is done?

- ~~Added `@sveltejs/kit` as dev dependency just for it's type `Config`~~ used updated `@sveltejs/vite-plugin-svelte` which provides a type `SvelteConfig` - fits the project better
- Removed unused dependencies
- Updated dependencies:
  - `@storybook/*` got defaulted to `^8.0.0` version, instead of `rc`
  - `@sveltejs/vite-plugin-svelte` to `^3.0.0`
  - `vite` to `5.0.0`
- Added `.svelte-kit/` to ignore list

### DIFF
--- a/src/config-loader.ts
+++ b/src/config-loader.ts
@@ -1,20 +1,20 @@
 // This file is a rewrite of `@sveltejs/vite-plugin-svelte` without the `Vite`
 // parts: https://github.com/sveltejs/vite-plugin-svelte/blob/e8e52deef93948da735c4ab69c54aced914926cf/packages/vite-plugin-svelte/src/utils/load-svelte-config.ts
-import { fileURLToPath, pathToFileURL } from 'url';
+import fs from 'node:fs';
+import { createRequire } from 'node:module';
+import path from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 
-import { createRequire } from 'module';
-import fs from 'fs';
 import { logger } from '@storybook/node-logger';
-import path from 'path';
-import { pathExists } from 'fs-extra';
+import type { Config } from '@sveltejs/kit';
 
 /**
  * Try find svelte config and then load it.
  *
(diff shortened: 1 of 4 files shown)


xeho91:
I'd update lockfile, but my current version of `pnpm` `v9` keeps upgrading the lockfile to new version - from `v6` to `v9`.

src/config-loader.ts
```diff
@@ -164,22 +163,21 @@ async function getConfigFilesUp() {
...
  * Directory in which to look for svelte files.
  *
- * @returns {Promise<string[]>}
+ * @returns
  * Returns an array containing all available config files.
  */
-async function getConfigFiles(lookupDir) {
-  /** @type {[string, boolean][]} */
-  const fileChecks = await Promise.all(
+async function getConfigFiles(lookupDir: string): Promise<string[]> {
```
JReinhold, line 172: We can convert this whole function to be sync now, right?

JReinhold (approved):
Minor nit, but otherwise looks good! ✨

Would you like to open a PR upgrading to pnpm 9?

- Adding the `packageManager` field to `package.json`
- Updating the lock-file
- Updating the pnpm versions in the GitHub workflows.

src/config-loader.ts
xeho91: Correct, no more promises.

shilman:
:rocket: PR was released in [`v4.1.3`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.1.3) :rocket:

PR #182, opened 2024-05-16
Fix dependency on `@storybook/node-logger` and `@storybook/client-logger`

ndelangen:
# This fixes:

The core consolidation project will cause unresolved packages to start failing storybook.
This package has such unresolved dependencies.

Mind you this would have already broken `pnp` mode and possible other strict dependency managers.
So it's a thing we want to fix, even if the core package consolidation project is delayed or cancelled.

1 solution would be to move these to regular dependencies, instead of devDependencies.
But with the core package consolidation project in mind, removing/reducing the dependency on those packages is a good move.

# What I did:

- I updated some devDependencies, because why not?
- I removed the need for `@storybook/node-logger`, I deemed it not important enough to keep.
- I removed the use of `@storybook/client-logger` in the piece of code that's shared between server & browser.

The dependency on `@storybook/node-logger` was missing completely!

`@storybook/client-logger` was being referenced in a piece of code used by both node & browser.
This is fine for browser, because of prebundling (globalized in both preview & manager), but this is really bad for node, because:
- the dependency is missing
- it's loading browser code into node.

# QA:

I've manually tested this by:
- generating a svelte sandbox
- changing the dependency version
- installing
- running build-storybook (was not working before this patch)

<details>
  <summary>📦 Published PR as canary version: <code>4.1.3--canary.182.cb1f688.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@4.1.3--canary.182.cb1f688.0
  # or 
  yarn add @storybook/addon-svelte-csf@4.1.3--canary.182.cb1f688.0
  ```
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,6 @@
   },
   "license": "MIT",
   "type": "module",
-  "main": "./dist/index.js",
-  "packageManager": "pnpm@9.1.1",
-  "types": "./dist/index.d.ts",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
@@ -26,6 +23,8 @@
     },
     "./package.json": "./package.json"
   },
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "files": [
     "dist/**/*",
     "README.md",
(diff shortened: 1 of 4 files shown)


JReinhold (approved):
LGTM!

shilman:
:rocket: PR was released in [`v4.1.3`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.1.3) :rocket:

PR #181, opened 2024-05-14
Experimental support for Svelte 5

xeho91:
> [!CAUTION]
**This is still a work in progress** 🚧.
Breaking changes may occur at any time, so do not use it for other purposes than experimenting yet.

## Credits

- @tsar-boomba and his PR #177 

---

## Additional goals

1. Simplify the usage - align closer to regular CSF format, without extras
1. Take advantage of the Svelte 5 features
1. Improve examples demonstration

### Resources

- Demonstrations in the official demo: https://github.com/storybookjs/svelte-csf-demo
- Discussion on official Storybook https://github.com/storybookjs/storybook/discussions/27092

> [!TIP]
**Call for contribution on the demo _(by add missing or improving examples)_ to help us find potential issues. Some of them we can fix the next major version** 🙏.

---

## Breaking changes

- Dropped support for `webpack` - See discussion https://github.com/storybookjs/storybook/discussions/27092
- Removed `<Meta />` component
- Removed `<Template />` component in favor of Svelte snippets
- `<Story />` components no longer has a prop `autodocs`, use `tags={['autodocs']}` instead
- `<Story />` components no longer has a prop `source`, we've revamped the behaviour
- Removed codebase with support for Storybook `< v7.4`

---

## New features

1. **Improved TypeScript experience** - `defineMeta({ ... })` - works in TypeScript and with JSDoc
1. Added a function `setTemplate` to allow adding a default `{#snippet children()}` to all of the `<Story>` in the stories file.
1. Added type helpers `Args<typeof Story>` and `StoryContext<typeof Story>` to get typings while creating Svelte v5 snippets outside the `<Story />`

## TODO

- [x] Refactor of `indexer.ts` - it uses Svelte AST from `svelte.compile` to extract the data required for indexer
- [x] Refactor of parser:
  - move from `estree-walker` to `zimmerframe`,
  - use `modernAst` option in the `svelte.compile`
  - fix issues with parsing description from comment,
  - remove codebase related to `<Meta>` and `<Template>` component
- [x] Improve naming
- [x] Improve type-safety on the entire codebase
- [x] Migrate `<Story>` to use Svelte 5 features
- [x] Migrate addon's components `context` to use Svelte 5 runes instead of `writable()`
- [x] Update examples in `stories/`
- [x] Remove unused dependencies 
- [x] Update tests
- Improve error messages
  - [x] compiler
  - parser
    - [x] analyse
    - [x] extract 
  - [x] runtime
[...]

PR #186, opened 2024-07-05
Restore & add support for legacy syntax

xeho91:
## Dependencies
Requires #181 to be merged first.

### Support for legacy components for backward compatibility:

1. [`<Meta>`](https://github.com/storybookjs/addon-svelte-csf/blob/main/src/components/Meta.svelte)
    - [x] codemod - `<Meta>` to `defineMeta` & insertion into `<script context="module">`
    - [x] tests

1. [`<Story>`](https://github.com/storybookjs/addon-svelte-csf/blob/main/src/components/Story.svelte)
   - [x] tests
   - [x] codemods:
      - [x] transform `let:args` & `let:context` directives to inner `children` snippet block
      - [x] transform `autodocs` prop to `tags={["autodocs"]}`
      - [x] transform `source` by removing when is a shorthand or to `parameters.docs.source.code`
      - [x] transform `template` prop to `children={<template identifier>}`
        
1. [`<Template>`](https://github.com/storybookjs/addon-svelte-csf/blob/main/src/components/Template.svelte)
   - [x] codemods - replace component with snippet block
   - [x] tests

### Add addon options - `StorybookAddonSvelteCsfOptions`

- [x] update addon's Vite plugin - add `pre-transform` hook
- [x] add option `supportLegacy`

### Deprecation warnings, errors, messages

- [x] document new errors 
- [x] Ensure deprecation warnings are visible enough while writing stories code:
      - writing the code _(with TypeScript support and marking `<Template>` as deprecated)_
- [ ] log deprecated warnings at runtime

### Update stories tests

- [x] Using `LegacyTemplate` component
- [x] Using `LegacyStory` component _(and its legacy props)_
- [x] Using `Meta` component

### Update indexer

- [x] Support indexing legacy `export const meta` syntax
- [x] Support indexing legacy `Meta` component

### TODO

- [ ] https://github.com/storybookjs/addon-svelte-csf/pull/186#discussion_r1685643713
- [x] https://github.com/storybookjs/addon-svelte-csf/pull/186#discussion_r1685644016
- [x] https://github.com/storybookjs/addon-svelte-csf/pull/186#discussion_r1683370952
- [x] https://github.com/storybookjs/addon-svelte-csf/pull/186#discussion_r1685313983

---

<details>
  <summary>📦 Published PR as canary version: <code>4.2.0--canary.186.ff8a94d.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@4.2.0--canary.186.ff8a94d.0
  # or 
  yarn add @storybook/addon-svelte-csf@4.2.0--canary.186.ff8a94d.0
  ```
</details>


xeho91:
@JReinhold 

I got tests working. I'm encountering the issue with Chromatic CI which I can't figure out. I don't believe it's related to my changes.

What's left is updating documentation. Most importantly change wording where we declared 'removed' legacy API to 'deprecated'. And also, highlighting usage of the option `legacyTemplate`.

And maybe improve test _(stories)_ for legacy API, because I got a feeling you don't want to provide examples for legacy API, right?

src/runtime/LegacyMeta.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
```
JReinhold, line 1: All these legacy components could just throw an error in their instance script right?
If they actually exist at runtime, it must be because the user hans't enabled the legacy flag right?

Perhaps a nice documented error.

src/utils/error/codemod/index.ts
```diff
@@ -0,0 +1,30 @@
...
+    filename: StorybookSvelteCSFError['filename'];
+    attribute: InvalidTemplateAttribute['attribute'];
+  }) {
+    super({ filename });
+    this.attribute = attribute;
+  }
+
+  template(): string {
+    return dedent`
+      'Story' component prop 'template' value must be a text (string) with a reference to existing '<Template>' component id.
```
JReinhold, line 25:
```suggestion
      'Story' component prop 'template' value must be a string with a reference to existing '<Template>' component id.
```

tests/stories/LegacyMeta.stories.svelte
```diff
@@ -0,0 +1,9 @@
+<script context="module">
+  import { Meta, Story } from '@storybook/addon-svelte-csf';
+
+  import LegacyMeta from './LegacyMeta.svelte';
+</script>
+
```
JReinhold, line 6: The legacy API has never worked without a `<Template>` call, so we should add one here, even though it technically works without it in v5.

tests/stories/LegacyStory.stories.svelte
```diff
@@ -0,0 +1,42 @@
...
+   * * _Italic_,
+   * * `Code`.
+   */
+  export const meta = {
+    title: 'LegacyStory',
+    component: LegacyStory,
+    tags: ['autodocs'],
+  } satisfies Meta<LegacyStory>;
+</script>
+
```
JReinhold, line 22: Same as above, needs a `<Template>` here.

tests/stories/LegacyStory.stories.svelte
```diff
@@ -0,0 +1,42 @@
...
+  <LegacyStory {...args} />
+</Story>
+
+<Story name="With autodocs" autodocs />
+
+<Story name="With source as shorthand" source />
+
+<Story name="With source as text" source="<LegacyStory>{'Hi'}</LegacyStory>">
+  <LegacyStory>{'Hi'}</LegacyStory>
+</Story>
```
JReinhold, line 42: Looks like the source snippet is wrong here. See:

image

tests/stories/LegacyTemplate.stories.svelte
```diff
@@ -0,0 +1,43 @@
...
+</Template>
+
+<Template id="rounded" let:args>
+  <p>Using rounded template</p>
+  <LegacyTemplate {...args} onclick={handleClick} rounded />
+</Template>
+
+<Story name="Default" />
+
+<Story name="Rounded" template="rounded" />
```
JReinhold, line 43: I think _this_ is causing the Chromatic failure. I expect you'll see the same if you build the Storybook locally and serve it, maybe even in dev mode too.

This is the raw story as opened in an iframe: https://6042cb92fd1bb200234586ee-poqxheiipz.chromatic.com/iframe.html?id=tests-legacytemplate--default&viewMode=story

First error:

image

Leading to this:

image

I think this is the mismatch in snippet reference that I've commented on earlier. The snippet will be named `rounded()`, but after the codemod the Story will reference it as `children={Rounded}` - and `Rounded` is coincidentally the export name of the story, so it's in fact not referencing the snippet but the story defined as `const Rounded = stories[...]` after post-transformation.

JReinhold (changes requested):
Great job! 👏 Added a lot of comments. The biggest thing here is probably that it doesn't handle stories written exclusively with the instance-script, which we need to support and test.

> And maybe improve test (stories) for legacy API, because I got a feeling you don't want to provide examples for legacy API, right?

Correct. I think the test stories you have for the legacy components are good - and they already caught a bug ;) - no need to add anything to the examples dir.
You could add a more "advanced" test story like the Button one we have in v4 currently: https://github.com/storybookjs/addon-svelte-csf/blob/main/stories/button.stories.svelte

tests/stories/LegacyTemplate.stories.svelte
xeho91: Haha, I shot myself in the foot with this one. I was troubled with how the hell, then I remember:

https://github.com/storybookjs/addon-svelte-csf/blob/89bd7f1375c391f76aec40390e66e3c0b73fbae2/src/compiler/pre-transform/codemods/legacy-story.ts#L231-L238


Thanks for this amazing work, and I'm going to make a note to myself. Ensure that runtime works by actually manually verifying the runtime 😅

src/runtime/LegacyMeta.svelte
xeho91: Very good point, you're right.

I think adding a checker inside the `index.ts` while importing the legacy components is hard _(no idea if there's a simple way to access addon options)_.
So I can do it while traversing on the Svelte AST inside the indexer parser and post-transform maybe _(because pre-transform will be skipped)_.

src/compiler/pre-transform/index.ts
```diff
@@ -0,0 +1,221 @@
...
+
+      next(state);
+    },
+
+    Root(node, context) {
+      const { fragment, module, ...rest } = node;
+      const { state, visit } = context;
+
+      if (!module) {
+        throw new MissingModuleTagError(filename);
```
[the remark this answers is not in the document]
xeho91: Oh, that complicates the work a bit, I think.

Two solutions:
1. Codemod add the `context="module"` to the `<script>` tag.
   Might be very dangerous operation, because there might be some logic created by the user unrelated to the Storybook usage. For example `onclick` snippets for button tests, etc.
2. Codemod will extract the most important pieces:
    1. import declarations for this addon - `"@storybook/addon-svelte-csf"`
    1. import declaration of the component _(optional)_
    1. `export const meta`
  ... and create a separate `<script context="module">` above the instance tag, with the extracted AST nodes from the instance tag.

src/compiler/pre-transform/index.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (node.context === 'module') {
+        next(state);
+      }
+    },
+
+    Program(_node, context) {
+      const { next, state } = context;
+
+      next(state);
+    },
```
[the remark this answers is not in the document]
xeho91: This instructs the AST walker - `zimmerframe` to go "deeper" in the tree. I'm not sure if this can be omitted, but I can test.

`Root` -> `module` _(`Script`)_ -> `Program` -> `nodes` _(declarations)_.

src/compiler/pre-transform/codemods/component-meta-to-define-meta.test.ts
```diff
@@ -0,0 +1,57 @@
+import pkg from '@storybook/addon-svelte-csf/package.json' with { type: 'json' };
+import type { Comment, Component } from 'svelte/compiler';
+import { print } from 'svelte-ast-print';
+import { describe, it } from 'vitest';
+
+import { transformComponentMetaToDefineMeta } from './component-meta-to-define-meta';
+
+import { parseAndExtractSvelteNode } from '#tests/extractor';
+
+describe(transformComponentMetaToDefineMeta.name, () => {
```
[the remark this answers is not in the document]
xeho91: 1. Done in fdebf5ca6486d5e08faef98fede1176156a1be2e
2. This one is tough.
   I'm not sure what should be expected outcome. We move this variable declaration to module tag?
   Currently it is transforming to:

   ```ts
   const { Story } = defineMeta({
      	component: WithParameters,
      	parameters: { ...parameters, baz: 'yes' }
   });
   ```
   which of course would produce an incorrect behaviour at runtime.
   
   This case might need more work. Because apart of `parameters` prop, there could be other properties _(from meta)_ being referenced in variables, e.g. `tags`, `play`, etc.

src/compiler/pre-transform/codemods/import-declaration.test.ts
```diff
@@ -0,0 +1,36 @@
+import pkg from '@storybook/addon-svelte-csf/package.json' with { type: 'json' };
+import type { ImportDeclaration } from 'estree';
+import { print } from 'svelte-ast-print';
+import { describe, it } from 'vitest';
+
+import { transformImportDeclaration } from './import-declaration';
+
+import { parseAndExtractSvelteNode } from '#tests/extractor';
+
+describe(transformImportDeclaration.name, () => {
```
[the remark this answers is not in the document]
xeho91: Correct, instance-level imports are not considered right now. I think I can extract the import declaration from instance, and put in the module tag.

tests/stories/LegacyStory.stories.svelte
xeho91: Fixed.

It happened because, I overlooked the that Svelte's compiler tries to compute what is inside the curly brackets in text expression value of attribute - in this case `{'Hi'}`

src/compiler/pre-transform/codemods/component-meta-to-define-meta.test.ts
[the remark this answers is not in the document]
JReinhold: I'm thinking either:

1. we attempt to move it to the module script. This can be tricky, because moving it could require moving _other_ variables as well that it references. This is very edgy though, so maybe not something we need to support.
2. or we detect the reference and log a warning to the user that the legacy API-usage couldn't be supported. Ignore the file completely, tell them to migrate the file to the new API.

src/compiler/pre-transform/index.ts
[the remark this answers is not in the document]
JReinhold: I think option 2 sounds best, if you think it's doable.

This is crucial though, as I think most usage of the v4 API is with the instance script today.

src/runtime/LegacyMeta.svelte
JReinhold: > no idea if there's a simple way to access addon options

It's doable, but not simple so let's try to avoid it.

> I can do it while traversing on the Svelte AST inside the indexer parser and post-transform maybe (because pre-transform will be skipped without having this option enabled).

Just making the components throw an error should be the simplest solution here right? Like

```svelte
<script>
  throw new Error("you have not enabled the legacy option...");
</script>
```

xeho91: That's much simpler, thanks for the hint!

Do we need to have test for it? No idea how we could do it in this repo _(with one Storybook config)_ and not breaking the CI(?)

JReinhold: Not easily no, I'm fine with this just having been tested manually.

src/compiler/pre-transform/index.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      </script>
+
+      <!-- This is a description for **default** Button. -->
+      <Story name="Default" />"
+    `);
+  });
+
+  it("replaces 'Meta' with inserting 'defineMeta' to module tag", async ({ expect }) => {
+    const code = dedent(`
+      <script context="module">
```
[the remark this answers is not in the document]
xeho91: Let me know if those tests got it covered:
- https://github.com/storybookjs/addon-svelte-csf/blob/7aae06f7838357739a73da1ba42074d22c275071/src/compiler/pre-transform/index.test.ts#L144-L158
- https://github.com/storybookjs/addon-svelte-csf/blob/7aae06f7838357739a73da1ba42074d22c275071/src/compiler/pre-transform/index.test.ts#L160-L186
- https://github.com/storybookjs/addon-svelte-csf/blob/7aae06f7838357739a73da1ba42074d22c275071/src/compiler/pre-transform/index.test.ts#L188-L220

If yes, mark this as resolved 🙏

src/compiler/pre-transform/codemods/component-meta-to-define-meta.test.ts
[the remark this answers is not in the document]
xeho91: 1. This is too much complex work. I can see it when I just finished initial support of moving from instance tag to module:
   - addon import declaration _(and transforming it)_
   - optional stories component import
   - export const meta _(and transforming it)_
   
   I'd have to walk on `ObjectExpression` on `export const meta` or `<Meta>` attributes _(props)_, and finding identifiers to each meta entry at the root level _(depth = 0)_. Now, imagine some very edge cases with depth level > 1 entries _(e.g. for `parameters`)_ located in the instance tag. 😅
   
2. I need to think on how to display this warning - as in how to detect references.

src/runtime/LegacyMeta.svelte
xeho91: In commit: 34d0eacc633683e850a663b1e65b7734f01658cd

I added this error, but after testing manually I don't think this was the right direction 😭 

The first error I got is:
```
storybook v8.2.5

info => Cleaning outputDir: storybook-static
info => Loading presets
info => Building manager..
info => Manager built (116 ms)
info => Building preview..
WARN 🚨 Unable to index ./tests/stories/LegacyWithInstance.stories.svelte:
WARN   SB_SVELTE_CSF_PARSER_EXTRACT_SVELTE_0001 (MissingModuleTagError): The file 'file:///Users/xeho91/Nextcloud/Projects/oss/addon-svelte-csf/tests/stories/LegacyWithInstance.stories.svelte'
WARN does not have a module context (<script context="module"> ... </script>).
WARN
WARN defineMeta(...) should be called inside a module script tag, like so:
WARN
WARN <script context="module">
WARN import { defineMeta } from "@storybook/addon-svelte-csf";
WARN
WARN const { Story } = defineMeta({});
WARN </script>
WARN
WARN More info: https://github.com/storybookjs/addon-svelte-csf/blob/v4.1.2/ERRORS.md#SB_SVELTE_CSF_PARSER_EXTRACT_SVELTE_0001
```

Which makes sense, because indexer parser is being run first. So I guess we have to move error to this parser, which would benefit us by giving us `filename` path.

xeho91: Your idea of using try catch and recognise known errors from indexer parser related to the legacy API _(https://github.com/storybookjs/addon-svelte-csf/commit/95a4becc4191af91533823a45a7047616d0c4b29)_ does the job - tested manually, thanks! 🙏

xeho91:
@JReinhold PR is ready for review.
Phew, there were some issues so hard to spot, because of missing type support from `svelte/compiler` on their AST nodes _([reference]( https://github.com/sveltejs/svelte/pull/12967))_

Two requests to put attention to:
1. If you could find any spots where we can emit warnings about using legacy syntax, it'll be awesome!
2. https://github.com/storybookjs/addon-svelte-csf/pull/186#discussion_r1685643713 This one is too hard for me to figure out at this moment. As discussed in private, is better left to user to fix very advanced cases on their own.

src/compiler/pre-transform/codemods/legacy-story.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    expect(print(transformLegacyStory({ node }))).toMatchInlineSnapshot(
+      `
+			"<Story name="Default" parameters={{
+				docs: { source: { code: "'<Button primary />'" } }
+			}} />"
+		`
+    );
+  });
+
+  it("'source' prop when is a text expression gets moved to 'parameters' prop", async ({
```
JReinhold, line 76: duplicate test name

```suggestion
  it("'source' prop when is a text expression and Story has children, gets moved to 'parameters' prop", async ({
```

src/compiler/pre-transform/codemods/legacy-story.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    expect(print(transformLegacyStory({ node }))).toMatchInlineSnapshot(`
+			"<Story name="Default">
+				{#snippet children(_args, context)}
+					<p>{context.id}</p>
+				{/snippet}
+			</Story>"
+		`);
+  });
+});
```
JReinhold, line 195: can we add a test more that uses both `args` and `context` directives?

src/compiler/pre-transform/codemods/legacy-story.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      <Story name="Default" autodocs />
+    `;
+    const node = await parseAndExtractSvelteNode<Component>(code, 'Component');
+
+    expect(print(transformLegacyStory({ node }))).toMatchInlineSnapshot(
+      `"<Story name="Default" tags={["autodocs"]} />"`
+    );
+  });
+
+  it("moving 'autodocs' prop doesn't break with existing 'tags' prop", async ({ expect }) => {
```
JReinhold, line 25: can we add two more tests to support what we discussed in https://github.com/storybookjs/addon-svelte-csf/pull/207#discussion_r1739989511:

1. `tags="singular"` is converted to `tags=["singular"]`
2. `autodocs tags="singular"` is converted to `tags=["autodocs", "singular"]`

src/compiler/pre-transform/codemods/component-meta-to-define-meta.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    const component = await parseAndExtractSvelteNode<Component>(code, 'Component');
+
+    expect(print(transformComponentMetaToDefineMeta({ component }))).toMatchInlineSnapshot(`
+      "const { Story } = defineMeta({
+      	component: WithParameters,
+      	parameters: { ...parameters, baz: 'yes' }
+      });"
+    `);
+  });
+});
```
JReinhold, line 111: can we add two more tests as per https://github.com/storybookjs/addon-svelte-csf/pull/207#discussion_r1739989511:

1. `<Meta tags="singular" />` -> `tags: ["singular"]`
1. `<Meta tags=["first", "second"] />` -> `tags: ["first", "second"]`

src/compiler/pre-transform/index.test.ts
```diff
@@ -0,0 +1,221 @@
...
+				import { defineMeta } from "@storybook/addon-svelte-csf";
+
+				/** This is a description for the **Button** component stories. */
+				const { Story } = defineMeta({ title: "Atoms/Button", component: Button });
+			</script>
+
+			{#snippet children(args)}
+				<Button {...args} />
+			{/snippet}
+			<Story name="Default" />"
```
JReinhold, line 95: this looks incorrect? `children` isn't referenced

src/compiler/pre-transform/index.ts
```diff
@@ -0,0 +1,366 @@
...
+        instance: transformedInstance,
+        module: transformedModule,
+      };
+    },
+
+    // NOTE: It walks either on module or instance tag
+    Script(node, context) {
+      const { next, state } = context;
+
+      state.currentScript = state.currentScript = node.context === 'module' ? 'module' : 'instance';
```
JReinhold, line 70:
```suggestion
      state.currentScript = node.context === 'module' ? 'module' : 'instance';
```

src/compiler/pre-transform/index.ts
```diff
@@ -0,0 +1,366 @@
...
+      const { declaration } = node;
+      const { state } = context;
+
+      if (
+        declaration &&
+        declaration.type === 'VariableDeclaration' &&
+        declaration.declarations[0].type === 'VariableDeclarator' &&
+        declaration.declarations[0].id.type === 'Identifier' &&
+        declaration.declarations[0].id.name === 'meta'
+      ) {
```
JReinhold, line 113: nit: early return

```suggestion
      if (!(
        declaration &&
        declaration.type === 'VariableDeclaration' &&
        declaration.declarations[0].type === 'VariableDeclarator' &&
        declaration.declarations[0].id.type === 'Identifier' &&
        declaration.declarations[0].id.name === 'meta'
      )) {
        return;
      }
```

src/index.test.ts
```diff
@@ -1,5 +1,5 @@
-import type { PlayFunctionContext } from '@storybook/types';
-import type { Component, ComponentProps, Snippet } from 'svelte';
+import type { PlayFunctionContext } from 'storybook/internal/types';
```
JReinhold, line 1: Let's revert this so we can keep supporting 8.0, if possible.

JReinhold (changes requested):
Very impressive work!

ERRORS.md
```diff
@@ -325,3 +330,37 @@ eg. `Some story name!!` will be converted to `SomeStoryName`.
 You can fix this collision by providing a unique `exportName` prop with`<Story exportName="SomeUniqueExportName" ... />`.
 
 See more in [the `exportName` API docs](./README.md#custom-export-name).
+
+## `CODEMOD`
+
+### `SB_SVELTE_CSF_CODEMOD_0001`
```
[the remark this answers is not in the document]
xeho91: Agreed, done

src/compiler/pre-transform/codemods/legacy-story.test.ts
xeho91: 2468f04cc9f9

src/compiler/pre-transform/codemods/legacy-story.test.ts
xeho91: I will, I need to create a codemod for this one.
I'll do it tomorrow, because it is a wrap for me today.

src/compiler/pre-transform/codemods/component-meta-to-define-meta.test.ts
xeho91: 09e65099b1fc810b3bc8bb5981133cfa41ac7fb6 and 3c361d487720e0bc213cf046bd218015cdbc7576 should cover it.

src/compiler/pre-transform/codemods/legacy-story.test.ts
xeho91: 09e65099b1fc810b3bc8bb5981133cfa41ac7fb6 and 3c361d487720e0bc213cf046bd218015cdbc7576 should cover it.

src/compiler/pre-transform/index.test.ts
xeho91: You're right, is totally incorrect. And I'm puzzled on what to do with this case.

Reminder:
1. Component `<Template>` attribute `id` is optional
2. `<Story />` component(s) by default were using the `default` template _(if existed)_, but the reference isn't visible on the stories file _(is a runtime mechanism)_

In this case, is easy to assume that the best solution to this problem is to achieve the following result:

```svelte
<script context="module">
import { defineMeta, setTemplate } from "@storybook/addon-svelte-csf";
/** This is a description for the **Button** component stories. */
const { Story } = defineMeta({ title: "Atoms/Button", component: Button });
</script>

<script>
setTemplate(template);
</script>

{#snippet template(args)}
   <Button {...args} />
{/snippet}

<Story name="Default" />

<Story name="Other" />
```

To create a codemod for this one complicates the logic a bit because:
1. I need to create a `<script>` instance
2. I need to add `setTemplate` to the import declaration for this addon inside the `<script module>` tag.

Do you see a simpler approach to this case?

JReinhold: 1. When you find a `<Template/>` without an explicit id, convert it to a snippet with a namespaced name so it won't collide with others, like `{#snippet sb_default_template_1()}`. Also give it a number, in the case where multiple id-less Templates are found, they can have number 1, 2 3... I don't know why anyone would have multiple id-less Templates, but just to be on the safe side?
2. When you find a `<Story/>` without an explicit template reference, make it reference the first default, eg. `children={sb_default_template_1}`

Then you don't need the `setTemplate` complexity at all?

xeho91: I did give it a try. cfc7beb82f63157e791470f73cfb1a5efd4e3c8b

Had to refactor a little bit by passing down the state from `zimmerframe` walk down to transformation function.

I only found one gotcha in `2.` point. Assuming someone did create a bizarre stories _(invalid, but AFAIK has a fallback mechanism behaviour)_, like this:

```svelte
      <Template let:context>
        <p>{context.args}</p>
      </Template>

      <Story name="Default" />

      <Template let:args>
        <Button {...args} />
      </Template>

      <Story name="NextOne" />
```

... which children should the second story `NextOne` refer to? I assumed it should match the latest one. And the tests shows current ouput:

```svelte
      {#snippet sb_default_template_1(_args, context)}
      	<p>{context.args}</p>
      {/snippet}

      <Story name="Default" children={sb_default_template_1} />

      {#snippet sb_default_template_2(args)}
      	<Button {...args} />
      {/snippet}

      <Story name="NextOne" children={sb_default_template_2} />
```

Hope that's ok. Regardless, this case is very edgy, hope it never happens. It was a little eye-opening, because I see why sometimes there's a need to be more conservative with API design. The previous API design is... too *flexible* 😅

xeho91: Argh, there's an edge case that comes with the currently implemented solution. I had to change the components order: c2b6b827913a

Explanation: CI was failing, because `<Story name="default" />` were transformed into `<Story name="default" children={sb_default_template_2} />`. And the reference `sb_default_template_2` isn't defined because the second `<Template>` has an `id` which is used as an identifier for its own transformation to snippet.

I could probably filter out '*id-less*' templates, so this bizarre edge case doesn't happen. But I'll put it on hold until I get to know your opinion on this solution.

JReinhold: Alright, I tested out what the behavior in v4 was - because we should just mimic that. With the follow stories:

```svelte
<Template let:args>
	<Button {...args}>
		First template
	</Button>
</Template>

<Story name="First Story" args={{ rounded: true }} />

<Template let:args>
	<Button {...args}>
		Second template
	</Button>
</Template>

<Story name="Second Story" source args={{ rounded: false }} />
```

You get these two, super weird stories:

image

image

So it seems that if you have multiple templates without `id`, they just combine into a single render?! I cannot fathom why users would ever want this explicitly, and they'd also notice this right away because the whole story renders double. So there's 0.0001 % chance this will ever happen. The same thing happens btw if multiple Templates have the same id.

We don't need to support multiple templates without id/with the same id. Ideally we'd throw a nice error if we detect this, but if that's too hard, then don't even waste time on it, just let it fail however it does naturally.

xeho91: Huh. It might be some issue with logic on the runtime scope of this addon.

Regardless, I see a way to throw an error if there's more than two *'id-less'* `<Template>` components and it certainly feels like an easier solution for this edge scenario.

xeho91: b1bf0c3b665e

PR #195, opened 2024-07-29
fix: simplifying destructing syntax to avoid vite optimisation issue

axel7083:
Following https://github.com/storybookjs/addon-svelte-csf/pull/181. We faced an issue (probably related to vite optimisation) as explained by @xeho91 in https://github.com/storybookjs/addon-svelte-csf/pull/181#issuecomment-2211041666

The problem was linked to the destructuring assignment syntax in the `src/runtime/Story.svelte` file with the props of the componet.

This PR remove the destructuring assignment in favour of a single variable, this fixing our compilation issue.

### DIFF
--- a/src/runtime/Story.svelte
+++ b/src/runtime/Story.svelte
@@ -55,8 +55,8 @@
     source?: never;
   } & ({ exportName: string } | { name: string });
 
-  const { children, name, exportName: exportNameProp, play, ...restProps }: Props = $props();
-  const exportName = exportNameProp ?? storyNameToExportName(name!);
+  const props: Props = $props();
+  const exportName = props.exportName ?? storyNameToExportName(props.name!);
 
   const extractor = useStoriesExtractor<TOverrideArgs, TCmp, TMeta>();
   const renderer = useStoryRenderer<TOverrideArgs, TCmp, TMeta>();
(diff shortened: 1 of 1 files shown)


xeho91:
I'm not sure if this the right solution to the problem.

AFAIK, `$props` rune return value object properties are [meant to be destructed]https://svelte-5-preview.vercel.app/docs/runes#$props)

I do remember about your issue, and yet I haven't been able to reproduce it on other projects than yours. I need to consider the possibility that maybe the cause of this issue lies somewhere else - the error message and the error stack trace could be misleading. I'll try to investigate again with `vite-plugin-inspect` tonight my time.

xeho91:
@axel7083 

I've got a possible lead.
I can't verify right now with `podman-desktop` repository, but if you could find a time...

I believe the error could be connected with `svelte-preprocess`.
Does the error still occur if you temporarily disable it inside `svelte.config.js`?

If yes, that'll be good news. Because that'll help us narrow the scope of where the issue occurs. And it might be easier to provide reproduction for Svelte maintainers.

axel7083:
> I believe the error could be connected with `svelte-preprocess`. Does the error still occur if you temporarily disable it inside `svelte.config.js`?

@xeho91 deleting the `svelte.config.js` file fixes the compilation issue, the problem seems to be linked to the `svelte-preprocess` library then. Thanks for narrowing down the problem

JReinhold:
Does this mean that we can close this, or is there still an actual issue hidden somewhere in the addon?

xeho91:
> Does this mean that we can close this, or is there still an actual issue hidden somewhere in the addon?

Yes, AFAIK the issue is related to vite configuration conflict between `vite-preprocess`, and `@sveltejs/vite-plugin-svelte`.

I we can't do anything on our side, nor I have idea how to resolve or provide a simple reproduction.

If it occurs again to someone else, and can provide a simple reproduction, then you'll have a better luck reporting it either to:
- https://github.com/sveltejs/vite-plugin-svelte
- or https://github.com/sveltejs/svelte-preprocess

In the light of how much I learned and know so far, I'm not sure which one is at fault.
Both repositories belong to same organisation, so they will be able to transfer issue in case. But first, a simple reproduction.

PR #198, opened 2024-08-05
Fix type errors due to imports from @storybook/types

### LINKED ISSUE #194: [Bug] Broken types - @storybook/types issue
### Describe the bug

At some points between 8.0.10 and 8.2.6, something changed with how types are handled internally in the storybook. More specifically, I've seen that generally types are now imported from `storybook/internal/types` rather than `@storybook/types`. The issue here is that it seems like the package `@storybook/types` is not a dependency of other storybook packages anymore, and addon-svelte-csf also doesn't depend on it (only through devDependencies, which are not propagated), so the types package not installed anymore, making any code that tries to import it fails.

More specifically I'm talking about this one line of code:
https://github.com/storybookjs/addon-svelte-csf/blob/35a9b514ea67b6a6a29aac31eeaaca98579738d2/src/index.d.ts#L2

1. Any project that uses Storybook 8.2.6 and addon-svelte-csf
2. Make sure to not manually depend on `@storybook/types`
3. Write a story
4. Seemengly random/weird errors:

```
Error: Object literal may only specify known properties, and 'args' does not exist in type 'Partial<StoryProps>'. (ts)
    const controls: Partial<StoryProps> = {
        args: {
            creating: false,
```

The problem is that, inside addon-svelte-csf, `StoryProps` extends `BaseAnnotations` from `@storybook/types`, which declares `args`, `argTypes` and many others. But, as the package is not installed, TypeScript doesn't find the interface it's supposed to extend, making it as if it doesn't extend anything at all.

### Workaround

Manually add `@storybook/types` as a dependency.

rChaoz:
Closes #194

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.22.6",
-    "@storybook/types": "^8.0.0-rc.2",
+    "@storybook/types": "^8.0.0",
     "dedent": "^1.2.0",
     "magic-string": "^0.30.1"
   },
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.22.6",
+    "@storybook/types": "^8.0.0-rc.2",
     "dedent": "^1.2.0",
     "magic-string": "^0.30.1"
   },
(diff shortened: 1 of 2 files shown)


JReinhold (changes requested):
This won't be compatible with Storybook versions below 8.2.

I think this should be done instead.

https://github.com/storybookjs/addon-svelte-csf/issues/194#issuecomment-2256314698

rChaoz:
@JReinhold like this?

package.json
```diff
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.22.6",
+    "@storybook/types": "^8.0.0-rc.2",
```
JReinhold, line 50: This is probably better

```suggestion
    "@storybook/types": "^8.0.0",
```

Also remember to update the lock file.

JReinhold (changes requested):
🙏 Thanks

package.json
rChaoz: Done

shilman:
:rocket: PR was released in [`v4.1.5`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.1.5) :rocket:

Issue #200, opened 2024-08-11
[Bug]: `v5` - using tags for individual stories causes error


bhvngt:
### Describe the bug

When `v5` of this addon is used, using tags for individual stories causes crash.

### Steps to reproduce the behavior

Here's the [reproduction](https://stackblitz.com/edit/sveltejs-kit-template-default-bxesk4) of this issue.

Note: For some unknown reason, stackblitz does not save .storybook folder. Hence, `storybook init` will have to be run to render storybook

Make following changes to `src/stories/Button.stories.svelte` file

```
<Story name="Secondary" tags={["!dev"]}/>
```

### Expected behavior

Expecting `Secondary` story to disappear from storybook sidebar.

### Screenshots and/or logs

![image](https://github.com/user-attachments/assets/e6e75ae1-c938-4a80-98c9-b0c8989da227)

xeho91:
Hey @bhvngt, thank you for testing out `v5` ❤️ and spotting this bug!
I confirm I was able to reproduce the issue.

@JReinhold You could extract more info from the console in web browser dev tools.

I handpicked this one:

```txt
runtime.js:4708 SB_PREVIEW_API_0006 (StoryIndexFetchError): Error fetching `/index.json`:

TypeError: Cannot read properties of undefined (reading 'type')

If you are in development, this likely indicates a problem with your Storybook process,
check the terminal for errors.

If you are in a deployed Storybook, there may have been an issue deploying the full Storybook
build.
    at ts.getStoryIndexFromServer (https://sveltejskittemplatedefaultbxes-hfev--6006--60a75bde.local-credentialless.webcontainer.io/sb-preview/runtime.js:8449:11)
    at async ts.initializeWithProjectAnnotations (https://sveltejskittemplatedefaultbxes-hfev--6006--60a75bde.local-credentialless.webcontainer.io/sb-preview/runtime.js:8432:15)
    at async ts.initialize (https://sveltejskittemplatedefaultbxes-hfev--6006--60a75bde.local-credentialless.webcontainer.io/sb-preview/runtime.js:8406:39)
```

Is it possible that it could be an upstream issue in `@storybook/preview-api`?

JReinhold:
I'm looking into this.

bhvngt:
Thanks @JReinhold

PR #206, opened 2024-08-27
v5: Fix tags being ignored

JReinhold:
Fixes https://github.com/storybookjs/addon-svelte-csf/issues/200

We can't and shouldn't "precombine" tags in the indexer, we should just pass them in, making sure they are in the right order. Storybook will internally combine the tags correctly, with the project-level tags.

Consider the following example:

- project-level: `['dev', 'test']` (this is the default)
- meta-level: `['!dev', '!test', '!autodocs']` (disable everything)
- story-level: `['test', 'autodocs']` (enable tests and docs for this specific story, still keeping sidebar disabled)

If the indexer combines the meta and story tags with `combineTags()`, they'll end up as `['test', 'autodocs']` (`'!dev'` will be omitted completely, because it's not enabled). Then this gets passed to Storybook, which combines them with project-level tags, resulting in `['dev', 'test', 'autodocs']`. This is wrong, `'dev'` should not be there as it was disabled at the meta-level, but Storybook doesn't know that because the indexer already omitted `'!dev'`.

If we instead don't combine anything and let Storybook do it _with the project level tags_, it will get the following series:

`['dev', 'test', '!dev', '!test', '!autodocs', 'test', 'autodocs']`. Parsing these sequentially will yield the correct result: `['test', 'autodocs']`

<details>
  <summary>📦 Published PR as canary version: <code>4.1.7--canary.206.e4feeae.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@4.1.7--canary.206.e4feeae.0
  # or 
  yarn add @storybook/addon-svelte-csf@4.1.7--canary.206.e4feeae.0
  ```
</details>

### DIFF
--- a/src/indexer/index.ts
+++ b/src/indexer/index.ts
@@ -83,7 +83,7 @@ export const indexer: Indexer = {
         exportName,
         name,
         title: metaTitle,
-        tags: combineTags(...metaTags, ...tags),
+        tags: [...metaTags, ...tags],
       } satisfies IndexInput;
     });
   },
--- a/tests/stories/Tags.stories.svelte
+++ b/tests/stories/Tags.stories.svelte
@@ -1,6 +1,11 @@
 <script context="module">
   import { defineMeta } from '@storybook/addon-svelte-csf';
 
+  /**
+   * Testing tags parsing in the addon's indexer.
+   * - In the sidebar, only the _"With Dev"_ story should be visible.
+   * - In docs, only the _"With Autodocs"_ story should be visible.
+   */
   const { Story } = defineMeta({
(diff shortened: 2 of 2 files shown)


xeho91 (approved):
Ahhh, I see. My misuse of `combineTags` was the culprit!
Thanks for explaining it so thoroughly 🙏

shilman:
:rocket: PR was released in [`v5.0.0-next.3`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v5.0.0-next.3) :rocket:

PR #207, opened 2024-08-30
Add support for story-level tags

### LINKED ISSUE #203: [Bug]: Story-level `tags` are ignored
I would like to exclude stories from appearing in the sidebar. I only want to have stories for interaction testing purposes, but not for users to use.

https://storybook.js.org/docs/api/csf#non-story-exports

```
const meta: Meta<typeof MyComponent> = {
  component: MyComponent,
  includeStories: ['SimpleStory', 'ComplexStory'], // 👈 Storybook loads these stories
  excludeStories: /.*Data$/, // 👈 Storybook ignores anything that contains Data
};
```

I am able to do:

```
<Story
  ...
  parameters={{
    docs: {
      // Disables the story from showing on the Docs page, but not in the Sidebar
      disable: true,
    },
  }}
/>
```

But this does not prevent it from still be selectable in the sidebar.

JReinhold:
Fixes #203

This PR adds support for story-level tags as described at https://storybook.js.org/docs/writing-stories/tags

Ie. `<Story name="..." tags={['!dev', 'autodocs']} />`

- Extracted tags from `Story` attributes
- Forwarded extracted stories to the index inputs
- Updated `StoryProps` types to support the `tags` prop
- Added test stories that demonstrates/tests the feature
- Upgraded the internal Storybook to ~8.1.0, because that's the first version that reliably works with tags as we know it today.

**If you want to use this feature today, make sure you're at least on Storybook version 8.1.0**

<details>
  <summary>📦 Published PR as canary version: <code>4.1.7--canary.207.1873b6d.1</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@4.1.7--canary.207.1873b6d.1
  # or 
  yarn add @storybook/addon-svelte-csf@4.1.7--canary.207.1873b6d.1
  ```
</details>


src/parser/extract-stories.test.ts
```diff
@@ -330,6 +340,131 @@ describe('extractSource', () => {
...
+  test('Only one story tag', () => {
+    expect(
+      extractStories(`
+        <script>
+          import { Story, Meta } from '@storybook/addon-svelte-csf';
+        </script>
+
+        <Meta title='test'/>
+
+        <Story name="Story1" tags="single" >
```
xeho91, line 398: This isn't following CSF format, right? Asking in case if we're supposed to preserve this behavior in `v5`

xeho91 (approved):

src/parser/extract-stories.test.ts
JReinhold: That's a good question. I think our legacy support _should_ support this, but the new API shouldn't, as this is not like regular CSF, and I only added it here because it's how Meta already works.

shilman:
:rocket: PR was released in [`v4.1.7`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v4.1.7) :rocket:

PR #215, opened 2024-10-17
fix(types): Allow `Snippet` mapper with primitives to match `undefined`

xeho91:
Resolves #214 

<details>
  <summary>📦 Published PR as canary version: <code>4.1.8--canary.215.f10113b.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @storybook/addon-svelte-csf@4.1.8--canary.215.f10113b.0
  # or 
  yarn add @storybook/addon-svelte-csf@4.1.8--canary.215.f10113b.0
  ```
</details>

PR #219, opened 2024-10-23
fix: Resolve existing type issues

xeho91:
## Things to note

- Removed first generic type parameter `TOverrideArgs` which was incomplete.
  I believe is better to focus on getting types to work correctly than adding a new incomplete feature - and we haven't got close on achieving it.
- **Used `Cmp` instead of `CmpOrArgs` as the base generic type parameter to simplify** the type system in this addon project
- use legacy syntax on tests and components related to legacy API
- replace `#` imports _(which uses map from `pkg.imports`)_ inside `./src/runtime/**/*` - `svelte2tsx` doesn't support it (yet?) - ❤️ @paoloricciuti for [detailed investigation](https://discord.com/channels/486522875931656193/1298412930089615360)

<details>
  <summary>📦 Published PR as canary version: <code>4.1.8--canary.219.161a0fb.0</code></summary>
  <br />

  ```bash
  npm install @storybook/addon-svelte-csf@4.1.8--canary.219.161a0fb.0
  # or 
  yarn add @storybook/addon-svelte-csf@4.1.8--canary.219.161a0fb.0
  ```
</details>


xeho91:
> replace # imports (which uses map from pkg.imports) inside ./src/runtime/**/* - svelte2tsx doesn't support it (yet?) - ❤️ @paoloricciuti for [detailed investigation](https://discord.com/channels/486522875931656193/1298412930089615360)

@paoloricciuti Should we create an issue on [sveltejs/language-tools](https://github.com/sveltejs/language-tools) regarding this one?

paoloricciuti:
> > replace # imports (which uses map from pkg.imports) inside ./src/runtime/**/* - svelte2tsx doesn't support it (yet?) - ❤️ @paoloricciuti for [detailed investigation](https://discord.com/channels/486522875931656193/1298412930089615360)
> 
> @paoloricciuti Should we create an issue on [sveltejs/language-tools](https://github.com/sveltejs/language-tools) regarding this one?

I'll investigate a bit better but from what I've seen the transformation is actually correct...the type is lost when TS emit the DTS.

examples/Button.stories.svelte
```diff
@@ -19,7 +20,7 @@
     component: Button,
     tags: ['autodocs'],
     args: {
-      children: 'Click me',
+      children: createRawSnippet(() => ({ render: () => 'Click me' })),
```
JReinhold, line 23: I don't think this is a good example we want to document, and I don't think it's needed because of line 42 below?

If it's to satisfy types, then I think we should use type modifications instead like `as`.

src/index.test.ts
```diff
@@ -1,45 +1,28 @@
...
-  // });
-
   it('works with provided meta entry "component" entry', () => {
     const { Story, meta } = defineMeta({
       component: Button,
       args: {
-        children: 'Click me',
+        children: createRawSnippet(() => ({
+          render: () => 'Click me',
+        })),
```
JReinhold, line 25: same comment, this shouldn't be necessary

src/runtime/LegacyStory.svelte
```diff
@@ -6,22 +6,21 @@
...
      * Which one of `<Template>` id should be used for rendering this story children?
+     * @deprecated use Svelte [snippets]() instead.
+     * @see
      */
     template?: string | null;
   };
 
-  let { args }: Props = $props();
-
+  let args: $$Props['args'];
```
JReinhold, line 23: I don't understand this change TBH

src/runtime/Story.svelte
```diff
@@ -47,20 +38,20 @@
...
      * Use `tags={['autodocs']}` instead.
      */
-    autodocs?: never;
+    autodocs?: boolean;
     /**
      * @deprecated
      * Use `parameters={{ docs: { source: { code: "..." } } }}` instead.
      */
-    source?: never;
+    source?: boolean | string;
```
JReinhold, line 46: These aren't correct, are they? We don't support `autodocs` and `source` attributes in the new `Story`, only in `LegacyStory`? I can't remember why we had them as type `never` instead of just removing them completely though. the "deprecated" tag is misleading, since these are actually removed, not just deprecated.

JReinhold (changes requested):

src/runtime/Story.svelte
xeho91: Yes, they're removed from the modern `Story` component.

The opt-in legacy support option handles them in `LegacyStory` with pre-transform plugin.

The reason I brought them back was to prevent users from having waterfall of type errors caused by usage of these props. It would occur while upgrading this addon in their existing project(s) to the next major version.

Hmm... Now given the second thought reading my reply, I should've put these type changes inside the `LegacyStory` instead. That sounds reasonable to you?

src/runtime/LegacyStory.svelte
xeho91: Apparently Svelte language tools are so smart and capable to recognize when legacy API is used.
And producing two different types (via import `Story|LegacyStory.svelte`.

While using `LegacyStory`, the type of this component was treated as modern one (`svelte@5` - `Component` - function).
Hence props types we're mismatched and couldn't use let:*` directives without type errors.
Examples of these errors can be found in the other [PR comment](https://github.com/storybookjs/addon-svelte-csf/pull/215#issuecomment-2418701326)  (which can be closed if we merge this one)


So, for language tools to treat this component as legacy `SvelteComponent` (class), I had to use the legacy syntax for defining props.

examples/Button.stories.svelte
xeho91: Changing to:

```diff
- children: createRawSnippet(() => ({ render: () => 'Click me' })),
+ children: 'Click me' as Snippet,
```

will produce the error:

```txt
Conversion of type 'string' to type 'Snippet<[]>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first. [2352]
```

So, the workaround is:

```diff
- children: 'Click me' as Snippet,
+ children: 'Click me' as unknown as Snippet,
```

Which is cumbersome. Shall we proceed with this?

xeho91: I provided another alternative, which can be reverted: 104d5e0cb9dc818c104834f31755e44eb59b4f41

paoloricciuti: Wait but do you want a snippet or a string? Because those are not interchangeable

JReinhold: Well, we want a string here, because we convert it to a snippet down in the template below:

https://github.com/storybookjs/addon-svelte-csf/pull/219/files/6b646f1b2e9a44c8e52814b23ae2d71f93797b07#diff-27706a42acb167034a3d3aa2f621b4a43e00b1bd177c0e87c2028fad8e14b398R41-R43

This procedure ensures you can easily modify the arg as a string in the Control panel, while it still being passed as a snippet to the component.
But it's slightly cumbersome, and I haven't found a way yet to support this pretty common use case better yet - ideas welcome!

JReinhold: @xeho91 your alternative looks cleaner, but now it doesn't allow you to modify it in Controls anymore though. but let's stick with it. 👍

src/runtime/LegacyStory.svelte
JReinhold: oh damn that's good. great find! ❤️

src/runtime/Story.svelte
JReinhold: hmm. 

> The reason I brought them back was to prevent users being flooded with type errors caused by usage of these props. It would occur while upgrading this addon in their existing project(s) to the next major version.

This seems reasonable to me.

> Hmm... Now given the second thought reading my reply, I should've put these type changes inside the LegacyStory instead. That sounds reasonable to you?

And this is also what would make sense to me - but doing this means that users will still get flooded with type errors, right? because no user is explicitly importing LegacyStory? I'm conflicted here. We could document that to get rid of these type errors, users need to do a search and replace for `Story -> LegacyStory` in their code base? But I'm not sure that even works at runtime, does it?

I'll defer to your best judgement here, I trust you.

JReinhold (approved):
Still have comments, but I'm approving, and letting you decide if you want to act on them or not. 🙂

src/runtime/Story.svelte
xeho91: > And this is also what would make sense to me - but doing this means that users will still get flooded with type errors, right? **because no user is explicitly importing LegacyStory**? I'm conflicted here. We could document that to get rid of these type errors, users need to do a search and replace for `Story -> LegacyStory` in their code base? But I'm not sure that even works at runtime, does it?

Hm, I'm trying to get how you see that they could be flooded by those type errors? I'm guessing there's a confusion between `Story` and `LegacyStory` _(which is exported and aliased as `Story` for legacy compatibility)_.

Below is legacy syntax:

```svelte
<script>
  import { Story } from '@storybook/addon-svelte-csf';
  //       👆 This one is the `LegacyStory` now, so its old props are left untouched, except for the deprecation remarks.
</script>
```

JReinhold: > which is exported and aliased as Story for legacy compatibility

Sorry, I totally forgot about that part. 😵‍💫

shilman:
:rocket: PR was released in [`v5.0.0-next.11`](https://github.com/storybookjs/addon-svelte-csf/releases/tag/v5.0.0-next.11) :rocket:

PR #226, opened 2024-10-31
fix: properly transform invalid identifiers

paoloricciuti:
This fix the transform to handle the case where the template name is an invalid identifier.

Before it was using `camelCase` (only for the `Story` actually so that was also a bug) but the problem with `camelCase` is that it can lead to the same name from different string. For example `Modal && Trigger` and `Modal || Trigger` have the same `camelCase`.

To fix this i used a small hash function (stolen from the svelte code base) so that if the template is not a valid identifier (which needs to be to be a `Snippet`) it convert it to something like `template_skajsh` (which we don't care too much except for stack traces).

I don't know if it's a solution that you like but it works.

### DIFF
--- a/src/compiler/pre-transform/codemods/legacy-story.ts
+++ b/src/compiler/pre-transform/codemods/legacy-story.ts
@@ -1,5 +1,3 @@
-import { camelCase } from 'es-toolkit/string';
-
 import {
   createASTArrayExpression,
   createASTAttribute,
@@ -11,6 +9,7 @@ import {
 } from '#parser/ast';
 import { InvalidTemplateAttribute } from '#utils/error/legacy-api/index';
 
+import { hashTemplateName } from '#utils/identifier-utils';
 import type { State } from '..';
 
 interface Params {
@@ -247,7 +246,7 @@ function templateToChildren(
     value: [
       createASTExpressionTag({
         type: 'Identifier',
-        name: camelCase(
+        name: hashTemplateName(
           value[0].type === 'Text'
             ? value[0].data
             : ((value[0].expression as ESTreeAST.Literal).value as string)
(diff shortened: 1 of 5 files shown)


xeho91:
🧡 

Apologies for missing the case where the transformation of the legacy template component `id` needs to be a valid identifier for snippet.

---

BTW, Jeppe. We could implement some formatting CI to this repo _(Prettier probably is the best choice since it supports Svelte, unlike Biome)_.

src/compiler/pre-transform/codemods/legacy-story.ts
```diff
@@ -11,6 +9,7 @@ import {
 } from '$lib/parser/ast.js';
 import { InvalidTemplateAttribute } from '$lib/utils/error/legacy-api/index.js';
 
+import { hashTemplateName } from '#utils/identifier-utils';
```
JReinhold, line 12:
```suggestion
import { hashTemplateName } from '$lib/utils/identifier-utils.js';
```

JReinhold (approved):
🙏

JReinhold:
@paoloricciuti do you know why GH Actions aren't being runned in this PR? Did you disable it in your fork?

paoloricciuti:
> @paoloricciuti do you know why GH Actions aren't being runned in this PR? Did you disable it in your fork?
> 
> (Gave you write-access to the repo so you don't need forks in the future)

Thanks @JReinhold ...no i did not disable anything at least not by choice...let me check

paoloricciuti:
image

It should be ok...if you want i can redo the changes in a branch of the repo

JReinhold:
> if you want i can redo the changes in a branch of the repo

Thanks that would be great, probably the easiest solution right now.

paoloricciuti:
Closing in favor of #246

PR #242, opened 2024-12-13
fix: Prevent parser indexer not letting other addon errors to throw

xeho91:
Following [thread](https://discord.com/channels/486522875931656193/1316869674608099430) on Discord.

The culprit was that `try { ... } catch(error) { }` did not allow other _(than those hand-picked ones)_ errors from this addon to throw.

### DIFF
--- a/src/indexer/index.ts
+++ b/src/indexer/index.ts
@@ -9,6 +9,7 @@ import {
 } from '#utils/error/parser/extract/svelte';
 import { LegacyTemplateNotEnabledError } from '#utils/error/legacy-api/index';
 import { NoDestructuredDefineMetaCallError } from '#utils/error/parser/analyse/define-meta';
+import { isStorybookSvelteCSFError } from '#utils/error';
 
 export const createIndexer = (legacyTemplate: boolean): Indexer => ({
   test: /\.svelte$/,
@@ -37,6 +38,11 @@ export const createIndexer = (legacyTemplate: boolean): Indexer => ({
       ) {
         const { filename } = error;
         throw new LegacyTemplateNotEnabledError(filename);
+      }
+
+      // WARN: We can't use instance of `StorybookSvelteCSFError`, because is an _abstract_ class :sob:
+      if (isStorybookSvelteCSFError(error)) {
+        throw error;
       }
 
       throw new IndexerParseError();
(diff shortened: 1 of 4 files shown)


src/utils/error.ts
```diff
@@ -178,3 +178,18 @@ export abstract class StorybookSvelteCSFError extends Error {
...
   }
 }
+
+// WARN: We can't use instance of `StorybookSvelteCSFError`, because is an _abstract_ class :sob:
+export function isStorybookSvelteCSFError(error: unknown) {
+  if (typeof error !== 'object' || error === null) {
+    return false;
+  }
+
+  for (const key of ['category', 'code', 'data', 'documentation', 'fullErrorCode', 'template']) {
```
JReinhold, line 188: this looks like a list we'll forget to update when we change the `StorybookError` fields later, so I see two alternatives:

1. just make it minimal, like only looking for `fullErrorCode`
2. Add an explicit field to `StorybookError` like `isStorybookError = true`. Then you also wouldn't need this utility at all, you could maybe just check that field directly when needed.

xeho91: I haven't thought of this, thanks!

This definitely simplifies the process and code: 3b5e794c033191938818a667bcbec6eb3f3a6092

src/indexer/index.ts
```diff
@@ -39,6 +40,11 @@ export const createIndexer = (legacyTemplate: boolean): Indexer => ({
         throw new LegacyTemplateNotEnabledError(filename);
       }
 
+      // WARN: We can't use instance of `StorybookSvelteCSFError`, because is an _abstract_ class :sob:
+      if (isStorybookSvelteCSFError(error)) {
+        throw error;
+      }
+
       throw new IndexerParseError();
```
dominikg, line 48: you might want to pass in cause here to help understand where it originated. completely swallowing the og error and throwing a non-descriptive one requires bringing a debugger to inspect... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

src/indexer/index.ts
```diff
@@ -39,7 +42,12 @@ export const createIndexer = (legacyTemplate: boolean): Indexer => ({
         throw new LegacyTemplateNotEnabledError(filename);
```
JReinhold, line 42: add the cause here too?

JReinhold (approved):

Issue #229, opened 2024-11-12
[Bug] Fails to load in Node 23 with `SB_CORE-SERVER_0002 (CriticalPresetLoadError)`


JReinhold:
> [!NOTE]
> This issue has a [temporary workaround](https://github.com/storybookjs/addon-svelte-csf/issues/229#issuecomment-2540390494)

### Describe the bug

When starting up a fresh Storybook project (both 8.3 and 8.4) with Svelte 5 and v5 of this addon, it fails to start with the error pasted below.

Downgrading Node to v22 fixes it. So does removing the addon and removing `.svelte` from the `stories` glob, which is why I've narrowed it down to this.

### Screenshots and/or logs

```
SB_CORE-SERVER_0002 (CriticalPresetLoadError): Storybook failed to load the following preset: ./.storybook/main.js.

Please check whether your setup is correct, the Storybook dependencies (and their peer dependencies) are installed correctly and there are no package version clashes.

If you believe this is a bug, please open an issue on Github.

SB_CORE-SERVER_0002 (CriticalPresetLoadError): Storybook failed to load the following preset: ./node_modules/@storybook/addon-svelte-csf/dist/preset.js.

Please check whether your setup is correct, the Storybook dependencies (and their peer dependencies) are installed correctly and there are no package version clashes.

If you believe this is a bug, please open an issue on Github.

Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import './node_modules/@storybook/addon-svelte-csf/dist/compiler/post-transform/story' is not supported resolving ES modules imported from ./node_modules/@storybook/addon-svelte-csf/dist/compiler/post-transform/index.js
    at finalizeResolution (node:internal/modules/esm/resolve:263:11)
    at moduleResolve (node:internal/modules/esm/resolve:932:10)
    at defaultResolve (node:internal/modules/esm/resolve:1056:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:650:12)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:599:25)
    at ModuleLoader.getModuleJobForRequire (node:internal/modules/esm/loader:349:53)
    at new ModuleJobSync (node:internal/modules/esm/module_job:338:34)
    at ModuleLoader.getModuleJobForRequire (node:internal/modules/esm/loader:409:11)
    at new ModuleJobSync (node:internal/modules/esm/module_job:338:34)
[...]

PR #243, opened 2024-12-13
Simplify imports

JReinhold:
Closes https://github.com/storybookjs/addon-svelte-csf/issues/229, implements https://github.com/storybookjs/addon-svelte-csf/issues/229#issuecomment-2540843640

- Add SvelteKit to correctly get the default `$lib` alias
- Move all files to `src/lib`
- Remove subpath imports and replace with `$lib` alias
- Add explicit `.js` extensions and `/index.js` to dir imports


JReinhold:
Need to work out the last details to silent SvelteKit errors during build.

PR #246, opened 2024-12-18
fix: properly transform invalid identifiers

paoloricciuti:
Alternative to #226 since that was created on a fork and apparently can't run actions 🤷🏼 

---

This fix the transform to handle the case where the template name is an invalid identifier.

Before it was using camelCase (only for the Story actually so that was also a bug) but the problem with camelCase is that it can lead to the same name from different string. For example Modal && Trigger and Modal || Trigger have the same camelCase.

To fix this i used a small hash function (stolen from the svelte code base) so that if the template is not a valid identifier (which needs to be to be a Snippet) it convert it to something like template_skajsh (which we don't care too much except for stack traces).

I don't know if it's a solution that you like but it works.


JReinhold (approved):
thanks a lot. ❤️

PR #266, opened 2025-01-17
Fix tags Story-level tags not having an effect in Vitest integration

### LINKED ISSUE #261: [Bug] Story-level tags are ignored in Vitest integration
### Describe the bug

When using the new Vitest integration via `@storybook/experimental-addon-test`, the way to skip specific stories from being tested is by removing the `'test'` tag. This is done by setting `'!test'` on a story:

```svelte
<Story name="My Story" tags={['!test']} />
```

Tags work fine in Storybook UI because the Svelte CSF indexer parses them via static analysis, but the indexer doesn't run in Vitest mode (it doesn't need to).

Instead, the Vitest plugin from `@storybook/experimental-addon-test` uses the `CsfFile` construct to statically analyse the (compiled) CSF file. But because tags aren't part of the compiled output of Svelte CSF, they aren't detected during static analysis.

This is only a problem for _story-level_ tags, meta or preview level tags are parsed correctly.

1. Have a basic Svelte CSF setup and add the addon with `npx storybook@next add @storybook/experimental-addon-test@next`
2. Run Vitest with something like `npm run test --project storybook`
3. See all stories testing
4. Add a `'!test'` tag to a story
5. See that it is not skipped, but it should be

### Additional context

`CsfFile` parses tags here: https://github.com/storybookjs/storybook/blob/next/code/core/src/csf-tools/CsfFile.ts/#L626-L631
And here are the tests that showcases what it supports: https://github.com/storybookjs/storybook/blob/next/code/core/src/csf-tools/CsfFile.test.ts/#L1278-L1375

The solution is to make sure tags are part of the compiled Svelte CSF.

Today we have:

```js
const __stories = createRuntimeStories(RequiredSnippet_stories, meta);

export default meta;

export const SomeStory = __stories["SomeStory"];
export const OtherStory = __stories["OtherStory"];
```

We could change that to one of the three:
```js
const __stories = createRuntimeStories(RequiredSnippet_stories, meta);

export const SomeStory = __stories["SomeStory"];
SomeStory.tags = ["some-tag", "other-tag"];
```

OR:

```js
const __stories = createRuntimeStories(RequiredSnippet_stories, meta);

export const SomeStory = {
  ...__stories["SomeStory"],
  tags: ["some-tag", "other-tag"]
};
```

OR:

```s
const __stories = createRuntimeStories(RequiredSnippet_stories, meta);

const SomeStoryTags = ["some-tag", "other-tag"];

export const SomeStory = {
  ...__stories["SomeStory"],
  tags: SomeStoryTags
};
```

All would work, we should just do what is simplest to implement.

xeho91:
Resolves #261


xeho91:
@JReinhold I only need an answer to [this question](https://github.com/storybookjs/addon-svelte-csf/issues/261#issuecomment-2597573760)

Technically speaking, it is possible to convert every story attribute to `Property`. But for now, I started with `tags`, because I don't know/remember if this would break anything.

JReinhold (approved):
Fantastic work. ❤️

PR #268, opened 2025-01-17
In Button example, fix `children` arg not being controllable

SteveALee:
I found the controls panel "children" text value was not working properly even though was declared in `argTypes`.


examples/Button.stories.svelte
```diff
@@ -20,6 +21,7 @@
     tags: ['autodocs'],
     args: {
       onclick: onclickFn,
+      children: 'Click me' as unknown as Snippet<[]>
```
SteveALee, line 24: Note sure if this is cleanst way but could find no examples of passing string as a children prop

JReinhold: I've changed to `any` instead

JReinhold (approved):
Thank you for this @SteveALee! ❤️ Sorry this took so long, I missed it!

Issue #270, opened 2025-01-21
[Bug] `mount()` in `play()` fails with `SB_PREVIEW_API_0012 (MountMustBeDestructuredError): Incorrect use of mount in the play function.`


JReinhold:
### Describe the bug

When using the `mount`-construct in a `play`-function as described in the docs:

https://storybook.js.org/docs/writing-tests/component-testing#run-code-before-the-component-gets-rendered

The Component Tests panel fails with the following message:

```
SB_PREVIEW_API_0012 (MountMustBeDestructuredError): Incorrect use of mount in the play function.

To use mount in the play function, you must satisfy the following two requirements: 

1. You *must* destructure the mount property from the `context` (the argument passed to your play function). 
   This makes sure that Storybook does not start rendering the story before the play function begins.

2. Your Storybook framework or builder must be configured to transpile to ES2017 or newer. 
   This is because destructuring statements and async/await usages are otherwise transpiled away, 
   which prevents Storybook from recognizing your usage of `mount`.

Note that Angular is not supported. As async/await is transpiled to support the zone.js polyfill. 

More info: https://storybook.js.org/docs/writing-tests/interaction-testing#run-code-before-the-component-gets-rendered

Received the following play function:
(storyContext) => {
        var _a;
        const delegate = (_a = storyContext.playFunction) == null ? void 0 : _a.__play;
        if (delegate) {
          return delegate(storyContext);
        }
        return play(storyContext);
      }
    at g.mount (http://localhost:6006/node_modules/.cache/storybook/1c3385a5d25e538d10b518b310c74d3ca2690b6aaffeadccd74da79736171f86/sb-vite/deps/storybook_internal_preview_runtime.js?v=0680545d:8004:19)
    at play (http://localhost:6006/tests/stories/Mount.stories.svelte:31:9)
    at storyObj.play (http://localhost:6006/node_modules/.cache/storybook/1c3385a5d25e538d10b518b310c74d3ca2690b6aaffeadccd74da79736171f86/sb-vite/deps/@storybook_addon-svelte-csf_internal_create-runtime-stories.js?v=0680545d:676:18)
    at http://localhost:6006/node_modules/.cache/storybook/1c3385a5d25e538d10b518b310c74d3ca2690b6aaffeadccd74da79736171f86/sb-vite/deps/storybook_internal_preview_runtime.js?v=0680545d:8005:60
[...]

PR #169, opened 2024-01-29
Fix play function not running in the component scope

j3rem1e:
@paoloricciuti  can you test this PR in your context please ? thanks !

PR #271, opened 2025-01-25
Fix destructured mount in play

SteveALee:
Addresses storybookjs/addon-svelte-csf#270

First time commit to storybook and attempt to fix the svelte `play({mount})` problem.

This was a bit tricky given steep learing curve as I went "round the houses" and I'm sure I missed loads of process essentials. If so some mentoring or pointers would help.

The problem was introduced with #169. I could not find a related issue to explain the problem but I guess it's pretty obvious. Hopefully the PR doesn't break anything.

Didn't find any relevent tests but guess some should be added for new code? Here's a story @JReinhold used:

```svelte
 <Story name="With Mount" play={async ({ mount }) => {
    console.log('pre 1');
    await mount();
    console.log('post 1');
}} />
```

See comments inline in the commits

### DIFF
--- a/src/runtime/create-runtime-stories.ts
+++ b/src/runtime/create-runtime-stories.ts
@@ -68,9 +68,21 @@ export const createRuntimeStories = (Stories: Component, meta: Meta<Cmp>) => {
        * The 'play' function should be delegated to the real play Story function
        * in order to be run into the component scope.
        */
-      function params(fn) { return fn.toString().match(/[^(]*\(([^)]*)/)?.slice(1) ?? [] }
-      const isMounting = params(storyObj.play).filter((p) => /\{\s*mount\s*\}/.test(p)).length != 0
-      storyObj.play = isMounting ? function ({ mount }) { return playDelegator(arguments[0]) } : (storyContext) => playDelegator(storyContext);
+      function params(fn) {
+        return (
+          fn
+            .toString()
+            .match(/[^(]*\(([^)]*)/)
+            ?.slice(1) ?? []
+        );
+      }
+      const isMounting =
+        params(storyObj.play).filter((p) => /\{.*,?\s*mount\s*,?.*\}/.test(p)).length != 0;
+      storyObj.play = isMounting
+        ? function ({ mount }) {
+            return playDelegator(arguments[0]);
+          }
+        : (storyContext) => playDelegator(storyContext);
       function playDelegator(storyContext) {
         const delegate = storyContext.playFunction?.__play;
 
@@ -80,7 +92,7 @@ export const createRuntimeStories = (Stories: Component, meta: Meta<Cmp>) => {
 
         // @ts-expect-error WARN: It should not affect user perspective- the problem lies in this addon's type `SvelteRenderer` missing type constrains or default generic parameter type
         return play(storyContext);
-      };
+      }
     }
 
     stories[exportName] = storyObj;
--- a/src/runtime/create-runtime-stories.ts
+++ b/src/runtime/create-runtime-stories.ts
@@ -82,7 +82,7 @@ export const createRuntimeStories = (Stories: Component, meta: Meta<Cmp>) => {
         ? function ({ mount }) {
             return playDelegator(arguments[0]);
           }
-        : (storyContext) => playDelegator(storyContext);
+        : playDelegator;
       function playDelegator(storyContext) {
         const delegate = storyContext.playFunction?.__play;
 
--- a/src/runtime/create-runtime-stories.ts
+++ b/src/runtime/create-runtime-stories.ts
@@ -68,7 +68,10 @@ export const createRuntimeStories = (Stories: Component, meta: Meta<Cmp>) => {
        * The 'play' function should be delegated to the real play Story function
        * in order to be run into the component scope.
        */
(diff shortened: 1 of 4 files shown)


src/runtime/create-runtime-stories.ts
```diff
@@ -68,7 +68,10 @@ export const createRuntimeStories = (Stories: Component, meta: Meta<Cmp>) => {
        * The 'play' function should be delegated to the real play Story function
        * in order to be run into the component scope.
        */
-      storyObj.play = (storyContext) => {
+      function params(fn) { return fn.toString().match(/[^(]*\(([^)]*)/)?.slice(1) ?? [] }
+      const isMounting = params(storyObj.play).filter((p) => /\{\s*mount\s*\}/.test(p)).length != 0
```
SteveALee, line 72: Don't think we can import the utils in core used to lok for `{mount}` so provided simplified equivalent

src/runtime/create-runtime-stories.ts
```diff
@@ -68,7 +68,10 @@ export const createRuntimeStories = (Stories: Component, meta: Meta<Cmp>) => {
...
+      storyObj.play = isMounting ? function ({ mount }) { return playDelegator(arguments[0]) } : (storyContext) => playDelegator(storyContext);
```
SteveALee, line 73: I used arguments as destructuring `...var` rest and then spread many miss any non own enumerable props.

SteveALee:
I also noticed, 

- a few typing errors, esp for `stories`
- Component is [imported in ](https://github.com/storybookjs/addon-svelte-csf/blob/8ba1b43bf6d6b34396c126cc37cea2b43e4768ed/src/runtime/create-runtime-stories.ts#L3) and is svelte 4 only and due to dissapear.
- TS [error hiding](https://github.com/storybookjs/addon-svelte-csf/blob/8ba1b43bf6d6b34396c126cc37cea2b43e4768ed/src/runtime/create-runtime-stories.ts#L81) is not needed - no such error occured for me

Hope that helps

SteveALee:
Things to look at when I have more time:

- optimise out the `: (storyContext) => playDelegator(storyContext)` to just `playDelegator` - should be OK as function statement or change to lambda
- What is `meta.play` and is the new code compatible with it.
- Can we pass earler destructured mount detection from core in prepareStory through to this phase and lose extra test on function.toString()?
- tests

SteveALee:
Here's a patch for use with [patch-package](https://www.npmjs.com/package/patch-package) in case anyone else needs this before merge

Save it as `@storybook+addon-svelte-csf+5.0.0-next.23`

```diff
diff --git a/node_modules/@storybook/addon-svelte-csf/dist/runtime/create-runtime-stories.js b/node_modules/@storybook/addon-svelte-csf/dist/runtime/create-runtime-stories.js
index 02dce1f..28381d4 100644
--- a/node_modules/@storybook/addon-svelte-csf/dist/runtime/create-runtime-stories.js
+++ b/node_modules/@storybook/addon-svelte-csf/dist/runtime/create-runtime-stories.js
@@ -55,7 +55,12 @@ export const createRuntimeStories = (Stories, meta) => {
              * The 'play' function should be delegated to the real play Story function
              * in order to be run into the component scope.
              */
-            storyObj.play = (storyContext) => {
+
+          // Steve: manually patched
+            function params(fn) { return fn.toString().match(/[^(]*\(([^)]*)/)?.slice(1) ?? [] }
+            const isMounting = params(storyObj.play).filter((p) => /\{.*,?\s*mount\s*,?.*\}/.test(p)).length != 0
+            storyObj.play = isMounting ? function ({ mount }) { return playDelegator(arguments[0]) } : playDelegator;
+            function playDelegator(storyContext) {
                 const delegate = storyContext.playFunction?.__play;
                 if (delegate) {
                     return delegate(storyContext);
```

SteveALee:
I've discovered a problem I can't resolve

If the default "configure your project" mdx is enabled then autoplay is defined for my mount test stories (despite this being a documented danger). For some reason then the built in [mount destructured test](https://github.com/storybookjs/storybook/blob/bb1e945844aaf7d2776981173dda01a5ede2a232/code/core/src/preview-api/modules/preview-web/render/StoryRender.ts#L308) is then false and boom!.

If the mdx is disabled (as I did) then all is fine. Autodoc tag makes no difference.

I don't understand why [this code](https://github.com/storybookjs/storybook/blob/bb1e945844aaf7d2776981173dda01a5ede2a232/code/core/src/preview-api/modules/preview-web/render/mount-utils.ts#L5) is returning false when sutoplay is in effect.  But it seems to be related to mdx and autoplay not this PR directly.

Something may have changed in my config as I'm sure I saw a comment in the MDX in red about setting autoplay to run the the play functions.

SteveALee:
Another issues, when I installed the new vitest component tests the "Run Tests" errors on the [main test](https://github.com/storybookjs/addon-svelte-csf/blob/edf3992d91efc10cb8df0167e09ec8fcfc883f08/tests/stories/Mountable.stories.svelte#L40) but the expect passes according tothe results panel. The same old mounted error occurs in the console (cmd line not F12)

At this point, while I'm happy to work oin it with some support, it's probably better to let others pick this PR up.

PR #285, opened 2025-03-05
fix: Temporarily disable save from UI feature

### LINKED ISSUE #240: [Bug]: Creating stories from Controls fails in Svelte CSF with `SyntaxError: Unexpected token, expected "}"`
### Describe the bug

When viewing a story written in [Svelte CSF](https://github.com/storybookjs/addon-svelte-csf), editing controls and pressing "Save", the UI shows `did not save, unknown error`, and the terminal shows an error.

This is because [the editing logic of controls](https://storybook.js.org/docs/essentials/controls#creating-and-editing-stories-from-controls) doesn't support Svelte CSF, only regular CSF, which is a completely different format, language even. 

1. Create a svelte-vite sandbox, all stories there should now be in Svelte CSF
2. Open Storybook dev
3. Edit a control and press save

### Screenshots and/or logs

Error in terminal:

```
Error writing to .\src\stories\Button.stories.svelte:
SyntaxError: Unexpected token, expected "}" (8:9)
    at constructor (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:14251:15)
    at Wl.raise (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:16508:54)
    at Wl.unexpected (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:16525:18)
    at Wl.expect (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:16789:27)
    at Wl.jsxParseExpressionContainer (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:18853:75)
    at Wl.jsxParseElementAt (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:18894:122)       
    at Wl.jsxParseElement (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:18915:32)
    at Wl.parseExprAtom (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:18924:37)
    at Wl.parseExprSubscripts (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:21426:78)      
    at Wl.parseUpdate (.\node_modules\.pnpm\@storybook+core@8.4.6\node_modules\@storybook\core\dist\babel\index.cjs:21416:45)
```

### Additional context

At the moment it's not feasible to implement Svelte CSF support in Create from Controls. The feature works by modifying the AST, and given it's a completely different AST, the logic would have to be completely different. It might even be impossible at the current stage, given the only tool that allows you to _print_ a Svelte AST to source is [`svelte-ast-print`](https://github.com/xeho91/svelte-ast-print) which lacks features such as TypeScript support.

We should instead disable the feature from the addon. That is done by setting a specific parameter, and the addon can set that parameter for all stories it touches, so it wouldn't affect stories written in regular CSF:

https://storybook.js.org/docs/essentials/controls#disablesavefromui

_Originally reported at https://github.com/sveltejs/cli/issues/339_

xeho91:
Resolves #240

### DIFF
--- a/src/compiler/post-transform/define-meta/insert-parameters.ts
+++ b/src/compiler/post-transform/define-meta/insert-parameters.ts
@@ -1,12 +1,13 @@
 import {
   findASTPropertyIndex,
   findPropertyDescriptionIndex,
+  findPropertyDisableFromUIIndex,
   findPropertyDocsIndex,
   findPropertyParametersIndex,
   getDescriptionPropertyValue,
   getDocsPropertyValue,
   getParametersPropertyValue,
-} from '$lib/compiler/post-transform/shared/description.js';
+} from '$lib/compiler/post-transform/shared/parameters.js';
 import { createASTObjectExpression, createASTProperty, type ESTreeAST } from '$lib/parser/ast.js';
 import type { SvelteASTNodes } from '$lib/parser/extract/svelte/nodes.js';
 import type { CompiledASTNodes } from '$lib/parser/extract/compiled/nodes.js';
@@ -21,6 +22,10 @@ interface Params {
 }
 
 /**
+ * This function inserts parameters to `defineMeta()`.
+ *
+ * ## Case 1
+ *
  * Attempt to insert JSDoc comment above the `defineMeta()` call.
  *
  * Before:
@@ -41,19 +46,22 @@ interface Params {
  *   },
  * });
  * ```
+ *
+ * ## Case 2
+ *
+ * Inserts https://storybook.js.org/docs/essentials/controls#disablesavefromui
+ * Reference: https://github.com/storybookjs/addon-svelte-csf/issues/240
+ * TODO: Restore this feature
  */
-export function insertDefineMetaJSDocCommentAsDescription(params: Params): void {
+export function insertDefineMetaParameters(params: Params): void {
   const { nodes, filename } = params;
-  const { compiled, svelte } = nodes;
-  const { defineMetaVariableDeclaration } = svelte;
-  const { leadingComments } = defineMetaVariableDeclaration;
 
-  if (!leadingComments) {
+  if (!nodes.svelte.defineMetaVariableDeclaration.leadingComments) {
     return;
   }
 
   const defineMetaFirstArgumentObjectExpression = getDefineMetaFirstArgumentObjectExpression({
-    nodes: compiled,
+    nodes: nodes.compiled,
     filename,
   });
 
@@ -68,6 +76,34 @@ export function insertDefineMetaJSDocCommentAsDescription(params: Params): void 
     );
   }
 
+  // Related to case 2
+  const disableSaveFromUIIndex = findPropertyDisableFromUIIndex({
+    filename,
+    node: defineMetaFirstArgumentObjectExpression,
+  });
+  // Related to case 2
+  if (disableSaveFromUIIndex === -1) {
+    getParametersPropertyValue({
+      filename,
+      node: defineMetaFirstArgumentObjectExpression,
+    }).properties.push(
+      createASTProperty('disableSaveFromUI', {
+        type: 'Literal',
+        value: true,
+      })
+    );
+  } else {
+    // WARN: We're overriding the existing value for user
+    getParametersPropertyValue({
+      filename,
+      node: defineMetaFirstArgumentObjectExpression,
+    }).properties[disableSaveFromUIIndex] = createASTProperty('disableSaveFromUI', {
+      type: 'Literal',
+      value: true,
+    });
+  }
+
+  // Related to case 1
   if (
     findPropertyDocsIndex({
       filename,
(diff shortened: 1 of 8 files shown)


src/compiler/post-transform/define-meta/insert-parameters.ts
```diff
@@ -68,6 +76,34 @@ export function insertDefineMetaJSDocCommentAsDescription(params: Params): void
...
+      filename,
+      node: defineMetaFirstArgumentObjectExpression,
+    }).properties.push(
+      createASTProperty('disableSaveFromUI', {
+        type: 'Literal',
+        value: true,
+      })
+    );
+  } else {
+    // WARN: We're overriding the existing value for user
```
xeho91, line 96: I'm not sure if this was the right decision. My gut tells me that we _should_ override for the user, because this feature will not work for the Svelte CSF addon, yet.

xeho91:
I wonder if we should also create a separate issue to restore this feature in the future. Whenever it is feasible to do so.

I mean, the part where Storybook addons are able to control/override what is being inserted into the original code of the stories file(s). This would need exploration on the Storybook core internals.

JReinhold:
Nice! I think we can simplify it. Parameters don't have to be statically analysable (and they can't be, because they're very dynamic), so maybe we should be able to get away with this by just setting the parameter at runtime here: https://github.com/storybookjs/addon-svelte-csf/blob/next/src/runtime/create-runtime-stories.ts#L51

Which seems a lot simpler to me than the current AST-based approach?

(I know this context is a little late to give now, sorry 🙇)

I'm fine with the refactor you did to the AST parsing for the parameters/description though, if you still want to keep that around.



> I wonder if we should also create a separate issue to restore this feature in the future. Whenever it is feasible to do so.

Yes we can do that.

xeho91:
> Nice! I think we can simplify it. Parameters don't have to be statically analysable (and they can't be, because they're very dynamic), so maybe we should be able to get away with this by just setting the parameter at runtime here: https://github.com/storybookjs/addon-svelte-csf/blob/next/src/runtime/create-runtime-stories.ts#L51

Hm, I haven't though of that 🤔.
I'm a little puzzled by what you meant that `parameters` _(I think we're talking about both in `defineMeta` and `<Story>`)_ are _very dynamic_?
In my mind, I'm stuck at example with inserting description either from JSDoc comment above `defineMeta` or HTML comment above every `<Story />`. I don't know if that's just one unique case.

> (I know this context is a little late to give now, sorry 🙇)

No worries ❤️!
I should've communicated firstly how I planned to solve this solution. It seems like I got too comfortable solving through AST transformation, which can lead to overcomplicating 😆

JReinhold:
I'm a little puzzled by what you meant that `parameters` _(I think we're talking about both in `defineMeta` and `<Story>`)_ are _very dynamic_? In my mind, I'm stuck at example with inserting description either from JSDoc comment above `defineMeta` or HTML comment above every `<Story />`. I don't know if that's just one unique case.

By parameters I mean `parameters` that you can set on either meta or story. They can be anything, users are free to set whatever they want, and however they want - such as functions that calls other stuff or imports variables. The point is, they are so dynamic that Storybook will most likely never try to statically analyse them for whatever reason (that's actually why we introduced tags, a statically analysable version of parameters).

My point is, that there is no need to set the parameter directly in the source using AST, it's totally fine to do it at runtime, which I'd expect to be simpler here.

The problem with the description is that to _get_ the description from the comments, _you_ have to do static analysis with the AST, but when you have that information you don't need to _set_ `parameters.docs.description` statically, it's fine to set it at runtime. But in this specific case with the description, given you're already doing AST stuff to read it, it might just be easier to also set it with the AST - that's up to you.

Am I making any sense? 😅

xeho91:
@JReinhold 

> Am I making any sense? 😅

Yes, thank you!

I've pushed the latest commit, with _mutating_ `meta` inside the `createRuntimeStories`. However, I think the only way I can test it is manually.
I've noticed that neither AST solution, nor mutating at runtime works. I've used `pnpm install` to make sure my dependencies are up to date.

I also tried manually adding `parameters.disableSaveFromUI: true` to `defineMeta`, as well `<Story parameters={{ disableSaveFromUI: true }} />`.

I'm missing something, do you see it?

JReinhold:
Haha, it's really subtle in the docs, but it's actually `parameters.controls.disableSaveFromUI` 😅

https://storybook.js.org/docs/essentials/controls#parameters-1
https://storybook.js.org/docs/essentials/controls#disable-creating-and-editing-of-stories

JReinhold (approved):

PR #291, opened 2025-04-19
ci(ESLint): Update ESLint & migrate to flat config & reconfigure

xeho91:
My editor was crashing non-stop due to `eslint-lsp` having issues with legacy config.
So, I took this opportunity to solve this case, by migrating to flat config and use recommended dependencies.


src/indexer/parser.ts
```diff
@@ -268,21 +268,19 @@ export async function parseForIndexer(
             const { name } = attribute;
 
             if (name === 'title') {
-              state.meta.title ===
-                getStringValueFromAttribute({
-                  component: node,
-                  node: attribute,
-                  filename,
-                });
+              state.meta.title = getStringValueFromAttribute({
```
xeho91, line 271: This one was wow - how come we didn't notice? 😅

JReinhold: 😬

src/runtime/emit-code.ts
```diff
@@ -84,7 +84,7 @@ export const generateCodeToEmit = ({ code, args }: { code: string; args: StoryOb
     // replace single arg references with their actual value,
     // eg. myProp={args.something} => myProp={"actual"}
     // or <h1>{args.something}</h1> => <h1>"actual"</h1>
-    .replace(/args(?:[\w\d_$\.\?\[\]"'])+/g, (argPath) => {
+    .replace(/args(?:[\w\d_$.?[\]"'])+/g, (argPath) => {
```
JReinhold, line 87: are you sure this is correct?

package.json
```diff
@@ -79,8 +81,8 @@
     "auto": "^11.1.6",
     "chromatic": "^11.28.2",
     "concurrently": "^8.2.2",
-    "eslint": "^7.32.0",
-    "eslint-plugin-storybook": "^0.11.3",
```
JReinhold, line 83: Why did you remove `eslint-plugin-storybook`? AFAIK it is compatible with flat-file config.

https://github.com/storybookjs/eslint-plugin-storybook#configuration-eslintconfigcmjs

xeho91: It's not the flat-file config compatibility that I was worried about.

I restored it. From a quick manual test, the rules of the Storybook plugin has no effects on files. Is it possible that I misconfigured it?

src/runtime/emit-code.ts
xeho91: I did a quick test: https://regex101.com/r/zmPmKz/1
There was a rule complaint: https://eslint.org/docs/latest/rules/no-useless-escape#rule-details

Unless there are edge cases, I can't be 100% sure it is correct 😅. We have tests which are passing, and I have confidence in ESLint maintainers - their [test for this rule](https://github.com/eslint/eslint/blob/40dd2998cedddb75e0514b2c5cc855293c85da41/tests/lib/rules/no-useless-escape.js#L12) is quite extensive.

JReinhold: SGTM

package.json
JReinhold: Ah, it probably only affects regular CSF, and the preview file. We dont have much content there, but let's keep it anyway.

JReinhold (approved):

PR #295, opened 2025-04-26
Breaking: Add support for `render` in `defineMeta`, replacing `setTemplate`-function

### LINKED ISSUE #259: [Request]: Replace new `setTemplate` API with `render` in `defineMeta`
With the release of Svelte 5.5.0, Svelte introduced "exportable snippets": https://svelte.dev/docs/svelte/snippet#Exporting-snippets

This essentially means that snippets can now be referenced from the module context, versus before where they could only be referenced from the instance context. Given that `defineMeta()` must always be used in the module context, this means that it's now possible to pass template snippets to `defineMeta`.

# Proposal

This allows us to get rid of the annoying [`setTemplate()` API](https://github.com/storybookjs/addon-svelte-csf/tree/next?tab=readme-ov-file#default-snippet), that is a shortcut for setting the a default template across all stories in a file. I dislike it for multiple reasons:

1. It requires you to add an instance-script, something that is rarely needed in stories file in general, so it's added verbosity.
2. Conceptually is does the same thing as Storybook's core `render` function on the meta-level, but it's doesn't look remotely like it.

Here is the API change that I propose:

```diff
<script module>
- import { defineMeta, setTemplate } from '@storybook/addon-svelte-csf';
+ import { defineMeta } from '@storybook/addon-svelte-csf';
  import MyComponent from './MyComponent.svelte';

  const { Story } = defineMeta({
    /* ... */
+   render: template
  });
</script>

-<script>
-  setTemplate(template);
-</script>

{#snippet template(args)}
  <MyComponent {...args}>
    ...
  </MyComponent>
{/snippet}

<Story name="With Default Template" />
```

# Rationale

Swapping `setTemplate` for a meta-level `render` makes the API almost exactly like regular CSF. This makes knowledge more transferable for users coming from other frameworks, and makes docs, tutorials, articles, etc. more useful in Svelte CSF.

# Implementation

A template snippet can't be the _actual_ `render` property of a story context, because the Svelte CSF sets the internal `StoryRenderer` component to that.

1. But we _should_ be able to pass it as a prop to `StoryRenderer` from within `createRuntimeStories` which has access to the `meta`.
https://github.com/storybookjs/addon-svelte-csf/blob/next/src/runtime/create-runtime-stories.ts/#L54-L61
2. I don't think we need to _remove_ the `render` from the meta, because all Svelte CSF stories have a `render` defined as per 1, which overrides anything in `meta` as per the inheritance principle

JReinhold:
Fixes #259 

This PR adds support for referencing a root-level snippet with the `render`-property in `defineMeta`, so it is closely aligned with regular CSF. This makes the `setTemplate`-function obsolete, and thus that API has been removed.

See rationale in #259.

Unfortunately we're running into this bug in the Svelte language tools: https://github.com/sveltejs/language-tools/issues/2653

The consequence of that is this:

image

There's not much we can do about it other than documenting it, which I have done. I don't think it's a blocker for this API, as it's still vastly better than `setTemplate`.

### `setTemplate`-function removed in favor of `render` in `defineMeta`

The `setTemplate`-function has been removed. Instead reference your default snippet with the `render`-property in `defineMeta`:

```diff
<script module>
- import { defineMeta, setTemplate } from '@storybook/addon-svelte-csf';
+ import { defineMeta } from '@storybook/addon-svelte-csf';
  import MyComponent from './MyComponent.svelte';

  const { Story } = defineMeta({
    /* ... */
+   render: template
  });
</script>

-<script>
-  setTemplate(template);
-</script>

{#snippet template(args)}
  <MyComponent {...args}>
    ...
  </MyComponent>
{/snippet}

<Story name="With Default Template" />
```

This new API achieves the same thing, but in a less verbose way, and is closer aligned with Storybook's regular CSF.

> [!IMPORTANT]
> There is currently a bug in the Svelte language tools, which causes TypeScript to error with `TS(2448): Block-scoped variable 'SNIPPET_NAMAE' used before its declaration.`. Until that is fixed, you have to silent it with `//@ts-ignore` or `//@ts-expect-error`. See https://github.com/sveltejs/language-tools/issues/2653


src/parser/extract/svelte/instance-nodes.ts
xeho91, line 1: Oh nice, some complexity gets removed

xeho91 (approved):
I couldn't find anything that you might have possibly forget.

I got to admin that I love how much complexity was removed!
Only sucks about this TypeScript error... perhaps you might be able to persuade Simon during the Svelte Summit to fix it 😏

PR #302, opened 2025-05-06
Fix types

### LINKED ISSUE #216: Type helper `Args` for `setTemplate` snippet doesn't satisfy component props interface
> [!NOTE]
> This case has a [workaround](https://github.com/storybookjs/addon-svelte-csf/issues/216#issuecomment-2456046896).

> Actually, while optional children work in that canary version, other props being required causes an error now.
> 
> Minimal example:
> 
> Button.stories.svelte:
> ```svelte
> <script context="module" lang="ts">
>   import { defineMeta, setTemplate, type Args } from "@storybook/addon-svelte-csf";
> 
>   import Button from "./Button.svelte";
> 
>   const { Story } = defineMeta({
>     component: Button,
>     tags: ["autodocs"],
>     args: {
>       children: "Click me",
>       variant: "filled",
>     },
>     argTypes: {
>       variant: { control: "select", options: ["filled", "outlined"] },
>       children: { control: "text" },
>     },
>   });
> </script>
> 
> <script lang="ts">
>   setTemplate(template);
> </script>
> 
> {#snippet template({ children, ...args }: Args<typeof Story>)}
>   <Button {...args}>{children}</Button>
> {/snippet}
> 
> <Story name="Primary" args={{ variant: "filled" }} />
> ```
> 
> Button.svelte:
> ```svelte
> <script lang="ts">
>   import type { Snippet } from "svelte";
> 
>   interface Props {
>     variant: "filled" | "outlined";
>     children?: Snippet;
>   }
> 
>   const { variant, children }: Props = $props();
> </script>
> 
> <button type="button" class={variant}>
>   {@render children?.()}
> </button>
> ```
> 
> Here's the error:
> 
> Image
>  

 _Originally posted by @tomasz13nocon in [#214](https://github.com/storybookjs/addon-svelte-csf/issues/214#issuecomment-2419422739)_

I need help on deciding how the type `Args` should reflect input from `defineMeta`.

Currently, there's a mismatch in the above example.
Component prop `variant` is **not optional**.
Once you hover on the `args` inside the `{#snippet template({ children, ...args }: Args<typeof Story>)}`, LSP type definition will show:

```ts
{
    variant?: "filled" | "outlined" | undefined;
}
```

### LINKED ISSUE #273: [Bug] [v5 next] Making component props Partial<T> introduces type errors
### Describe the bug

I want (the option for) args to be typed as they are on the component, or at least as I passed them to the story. addon-svelte-csf seems to [make the story props optional](https://github.com/storybookjs/addon-svelte-csf/blob/7bcba35c517654cfd9133a9c933b9e2436fdc8fc/src/runtime/Story.svelte#L11) using [Partial](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype). This results in type errors when plumbing my component props through `Story` args, ie:

```svelte
<Story name="MyStory" args={{...storyArgs}}>
	{#snippet children(args)}
		<MyComponent {...args} /> <!-- error: Type { foo?: ... | undefined } is not assignable to type `$$ComponentProps` -->
	{/snippet}
</Story>

```

Currently, this means I have to either:
- cast each arg to the correct type
- not plumb my args through the story, giving up access to the storybook controls feature along the way

The [readme for v5 #typescript](https://github.com/storybookjs/addon-svelte-csf/tree/next?tab=readme-ov-file#typescript) states:

````md
If you need to customize the type of the args,
you can pass in a generic type parameter to `defineMeta`
that will override the types inferred from the component:

```ts
const { Story } = defineMeta<{ anotherProp: boolean }>( ... );
```
````

However, the [type parameter for defineMeta needs to extend Cmp](https://github.com/storybookjs/addon-svelte-csf/blob/7bcba35c517654cfd9133a9c933b9e2436fdc8fc/src/index.ts#L17) and [Cmp is defined svelte's `Component<any>`](https://github.com/storybookjs/addon-svelte-csf/blob/7bcba35c517654cfd9133a9c933b9e2436fdc8fc/src/types.ts#L10). Notably, this is not the same as the type of `args` as described in the docs, so either this is a code bug or a doc bug.

1. Go to [`examples/Button.stories.svelte`](https://github.com/storybookjs/addon-svelte-csf/blob/7bcba35c517654cfd9133a9c933b9e2436fdc8fc/examples/Button.stories.svelte#L18)
2. Add a type parameter to defineMeta() that matches a prop for the button
3. Observe type error

### Expected behavior

My expectation is that I would be able to specify just the component props as the type parameter as described in the readme, and that the type parameter would be used to type the story args.

In place of that, I'd also be happy with something like a `strict` mode for the Story that doesn't mark the args as optional/`Partial` so that my stories enforce the types expected by the component.

Really I just want some mechanism to have my $$ComponentProps type be unaltered in my story args. Whether that is something like a `strict` option, explicitly setting a type parameter on defineMeta(), or some other mechanism is not important to me.

### Screenshots and/or logs

If applicable, add screenshots and/or logs to help explain your problem.

### Environment

 - OS: MacOS
 - Node.js version: 20.17.0
 - NPM version: 11.0.0
 - Browser (if applicable): n/a
 - Browser version (if applicable): n/a
 - Device (if applicable): n/a

### Additional context

Let me know if a minimal repro of this would be helpful and I will create one. Bonus points if you have a template I can use :)

### LINKED ISSUE #289: [Bug] Can't define additional args in 5.0.0-next.28
### Describe the bug

The docs mention the following was possible:

```ts
const { Story } = defineMeta<{ anotherProp: boolean }>( ... );
```

However, I get the following error:

```
ts: Type '{ content: string; }' does not satisfy the constraint 'Component<any, {}, string>'.
  Type '{ content: string; }' provides no match for the signature '(this: void, internals: Brand<"ComponentInternals">, props: any): { $on?(type: string, callback: (e: any) => void): () => void; $set?(props: Partial<any>): void; }'.
```

The code I use:

```ts
const { Story } = defineMeta<{ content: string }>({
  title: "Design System/Button",
  component: Button,
  tags: ["autodocs"],
  argTypes: {
    content: { control: "text" },
    size: { control: "select" },
    type: { control: "select" },
  },
});
```

The `Button` component contains a default Svelte 4 slot, but for Storybook I want people to be able to change the button's content in the Storybook UI by setting `content`.

See code above.

### Expected behavior

All args but `content` are inferred from the component props and `content` is accepted as an additional arg.

### Screenshots and/or logs

N/A

### Environment

 - OS: MacOS
 - Node.js version: 22.14.0
 - NPM version: 10.9.2
 - Browser (if applicable): N/A
 - Browser version (if applicable): N/A
 - Device (if applicable): N/A
 - Storybook: 8.6.12

### Additional context

We use Svelte 5, but components are written with Svelte 4 syntax, e.g. in the `Button`'s case, we use a `<slot />` for the content.

JReinhold:
This PR improves the type system for `defineMeta` and `Story` significantly.

Fixes #216
Fixes #273

`args` in the `template` snippet is now correctly inferred to be the component's props instead of partial args. A small nuance here, is that it means you have to _omit_ `children` from `args` before spreading it in, as otherwise Svelte will complain that you both have `children` as prop and as content:

```svelte
<!--                👇 you need to omit 'children' from args to satisfy Svelte's constraints -->
{#snippet template({ children, ...args }: Args, context: StoryContext<typeof Layout>)}
  <MyComponent {...args}>
    {children}
    {#snippet footer()}
      {args.footer}
    {/snippet}
  </MyComponent>
{/snippet}
```
 
Fixes #289

Instead of being able to pass in a generic to `defineMeta`, the type of args is automatically inferred from the snippet you pass to the `render`-property. That way, you type the parameters of your template snippet instead, and that will be used to define the `args`-types for meta and Stories. If you don't want to the `render`-property, you can instead type the parameters directly in each `template`-snippet within each `Story`.

See updates to Readme and the new `Types.stories.svelte` example stories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,9 +203,6 @@ Stories can still override this default snippet using any of the methods for def
 > [!NOTE]
 > Svelte has the limitation, that you can't reference a snippet from a `<script module>` if it reference any declarations in a non-module `<script>` (whether directly or indirectly, via other snippets). See [svelte.dev/docs/svelte/snippet#Exporting-snippets](https://svelte.dev/docs/svelte/snippet#Exporting-snippets)
 
-> [!IMPORTANT]
-> There is currently a bug in the Svelte language tools, which causes TypeScript to error with `TS(2448): Block-scoped variable 'SNIPPET_NAMAE' used before its declaration.`. Until that is fixed, you have to silent it with `//@ts-ignore` or `//@ts-expect-error`. See https://github.com/sveltejs/language-tools/issues/2653
-
 #### Custom export name
 
 Behind-the-scenes, each `<Story />` definition is compiled to a variable export like `export const MyStory = ...;`. In most cases you don't have to care about this detail, however sometimes naming conflicts can arise from this. The variable names are simplifications of the story names - to make them valid JavaScript variables.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -170,17 +170,17 @@ packages:
     resolution: {integrity: sha512-+562v9k4aI80m1+VuMHehNJWLOFjBnXn3tdOitzD0il5b7smkSBal4+a3oKiQTbrwMmN/TBUMDvbdoWDehgOww==}
 
   '@asamuzakjp/css-color@3.1.7':
-    resolution: {integrity: sha512-Ok5fYhtwdyJQmU1PpEv6Si7Y+A4cYb8yNM9oiIJC9TzXPMuN9fvdonKJqcnz9TbFqV6bQ8z0giRq0iaOpGZV2g==}
+    resolution: {integrity: sha512-Ok5fYhtwdyJQmU1PpEv6Si7Y+A4cYb8yNM9oiIJC9TzXPMuN9fvdonKJqcnz9TbFqV6bQ8z0giRq0iaOpGZV2g==, tarball: https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-3.1.7.tgz}
 
   '@auto-it/all-contributors@11.3.0':
-    resolution: {integrity: sha512-2d9y9P5mZoqrqkvIHfQmZ57sLS6vNhLW3bilqPbWW+hzzeJeGIx5cKLIHG+RfuDCeP9d3A3Ahnj7ilIMMUVENA==}
+    resolution: {integrity: sha512-2d9y9P5mZoqrqkvIHfQmZ57sLS6vNhLW3bilqPbWW+hzzeJeGIx5cKLIHG+RfuDCeP9d3A3Ahnj7ilIMMUVENA==, tarball: https://registry.npmjs.org/@auto-it/all-contributors/-/all-contributors-11.3.0.tgz}
 
   '@auto-it/bot-list@11.3.0':
     resolution: {integrity: sha512-+izoqAyOSiDVt3WcjVkSvLBV9c82VXLSf3oSWWcCeoxW/YDQ2AoInQ3M3EEyuBP+Yw9KQwGTTYHqpR7ZFkZpDQ==}
     engines: {node: '>=10.x'}
 
   '@auto-it/conventional-commits@11.3.0':
-    resolution: {integrity: sha512-+1j2Yz8SoyxV+ioeivT8GIYLSlegfvoV36OfI4cphwGB35RyuDAG58340ymhW0v7I3QWHCKSLGTagyJTwcpqoA==}
+    resolution: {integrity: sha512-+1j2Yz8SoyxV+ioeivT8GIYLSlegfvoV36OfI4cphwGB35RyuDAG58340ymhW0v7I3QWHCKSLGTagyJTwcpqoA==, tarball: https://registry.npmjs.org/@auto-it/conventional-commits/-/conventional-commits-11.3.0.tgz}
 
   '@auto-it/core@11.3.0':
     resolution: {integrity: sha512-3i7ooAhQJulVDG3gmdOioTXLhpFoS75Z/OsLV8ZkrEaEH/sfxlslqFx20VjWva7gMLl2iO8IjbRnlLhkXy5geg==}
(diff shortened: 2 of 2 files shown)


src/types.ts
```diff
@@ -1,26 +1,22 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import type {
   ComponentAnnotations as BaseComponentAnnotations,
   StoryAnnotations as BaseStoryAnnotations,
   StoryContext as BaseStoryContext,
   WebRenderer,
 } from 'storybook/internal/types';
 import type { Component, ComponentProps } from 'svelte';
-import type { SetOptional, Simplify } from 'type-fest';
```
xeho91, line 8: I think we can remove `type-fest` as `devDependency` now.

xeho91: [487c97c](https://github.com/storybookjs/addon-svelte-csf/pull/302/commits/487c97c47fb36c1c382b4ab1a875342d20ac566d)

xeho91: Never mind, I didn't notice it's being used in new stories `Types.stories.svelte`

examples/Templating.stories.svelte
```diff
@@ -15,13 +15,11 @@
      * reference any root-level snippet, for that snippet to be the fallback snippet,
      * that is used in any story without explicit template.
      */
-    //@ts-expect-error TS does not understand that the snippet is defined before this call
```
xeho91, line 18: Ah, finally fixed by upstream! 🧡

tests/stories/Example.svelte
```diff
@@ -1,9 +1,11 @@
 <script lang="ts">
+  import type { Snippet } from 'svelte';
   import type { HTMLButtonAttributes } from 'svelte/elements';
 
-  interface Props extends HTMLButtonAttributes {
+  type Props = {
```
xeho91, line 5: Was the reason for this change _(from `interface` to `type`)_ to reduce the amount of props being inferred?

tests/stories/Example.svelte
```diff
@@ -1,9 +1,11 @@
...
     rounded?: boolean;
-  }
+    children?: Snippet;
+  } & Pick<HTMLButtonAttributes, 'onclick' | 'onmouseenter' | 'onmouseleave'>;
```
xeho91, line 8: Nit: This probably can be simplified to this:

```suggestion
  } & Pick<HTMLButtonAttributes, 'children' | 'onclick' | 'onmouseenter' | 'onmouseleave'>;
```

tests/stories/Example.svelte
```diff
@@ -1,9 +1,11 @@
 <script lang="ts">
+  import type { Snippet } from 'svelte';
```
xeho91, line 2:
```suggestion
```

Follow up to:

tests/stories/Example.svelte
JReinhold: you're right, after I fixed all the types, I didn't actually need to narrow it down anymore. reverted.

PR #321, opened 2025-07-03
Fix raw code not being injected with Svelte v5.35.1+

### LINKED ISSUE #320: [Bug] Code snippet generation is broken with Svelte 5.35.1 and up
### Describe the bug

In Svelte [v5.35.1](https://github.com/sveltejs/svelte/releases/tag/svelte%405.35.1), https://github.com/sveltejs/svelte/pull/16255 was introduced that made some changes to the compiled output of a Svelte component, that broke how we post-transform that output.

We try to inject additional props to the `Story` component, that contains the raw source code, so that can be displayed at runtime. This transformation snapshot from Vite plugin inspect shows what's _supposed_ to happen - it's from 5.35.0:

Image

But now that the structure has changed, this transformation doesn't take place. This is from 5.35.1:

Image

before:

```js
Story(node_3, { name: 'With mainFontSize' });
```

after:

```js
$.add_svelte_meta(() => Story(node_3, { name: 'With mainFontSize' });
```

We should find another way to add that "raw source code" information at runtime, relying as little as possible on the structure of the compiled Svelte component as possible, as that is understandably brittle.

The code that looks for that `Story()` call is at https://github.com/storybookjs/addon-svelte-csf/blob/4e1a667d6ae4634fffde1372b1df06ce2b5fbee4/src/parser/extract/compiled/stories.ts#L14-L33

Potentially a shorter term solution is to modify that code to both work with pre and post v5.35.1.

1. Use the addon with Svelte v5.35.1 or above
2. Open an auto docs view of a story
3. Expand the "code" view of any of the displayed stories
4. See the source snippet in there being very wrong, because it's based on wrong data

JReinhold:
Fixes #320

Thanks to @paoloricciuti for the idea.

I also had to throw in _a bunch_ of `any` and stuff to fix type errors. Probably upgrading Svelte introduced proper upstream AST typing, and our current type system doesn't handle that too well. It works.


src/parser/extract/compiled/stories.ts
```diff
@@ -23,6 +23,8 @@ export async function extractStoriesNodesFromExportDefaultFn(params: Params) {
 
       if (node.callee.type === 'Identifier' && node.callee.name === storyIdentifier.name) {
         state.push(node);
+      } else {
+        context.next();
```
JReinhold, line 27: This is the actual fix for the issue

PR #328, opened 2025-09-19
Use correct import path for `storybook/preview-api`

ndelangen:
Related: https://github.com/storybookjs/storybook/pull/32507


JReinhold (changes requested):
This is a breaking change, because it means the addon doesn't support SB 8 anymore.
I'm not against taking that route, but it means that we need to do more work here to reflect that, and there are also other breaking changes that we've wanted to do for this addon's next major release.

